PR #32, opened 2022-05-18
Feature/task system

bomoko:
This PR introduces an initial pre/post-rollout task run command. Essentially it is supposed to bring us to parity with the existing build deploy bash, simply reading tasks out of a .lagoon.yml, grabbing (or scaling up) the appropriate pods, and exec'ing the task.

It's called like so

`lagoon-build tasks pre-rollout|post-rollout`

For the most part it sticks to the logic of the existing exec calls - checking for the existence of pods, scaling up, waiting, etc.

It does introduce a simple extension and DSL for `.lagoon.yml` - namely, as part of a function you can now specify a "when" clause - the "when" expects the expression contained within to evaluate to a boolean, and allows you to run simple comparisons against a set of environment variables - those given in the build pod itself, those passed in via LAGOON_PROJECT_VARIABLES, and those passed in via LAGOON_ENVIRONMENT_VARIABLES (these have increasing order of precedence - a variable occuring in ENVIRONMENT has precedence over the same name in PROJECT, and the same holds for the build pod).

For the most part, this is a simple language described here -> https://github.com/PaesslerAG/gval/blob/master/README.md

It has the following extensions though
* a function "withDefault" which can be used to protect a "when" clause from undefined variables (which cause panic) - eg, `when: withDefault("NONEXISTENTVARIABLE", false) == true`
* a function "exists" which simply returns true/false if a variable exists in the environment or not

These two functions exist for the moment to make handling undefined variables somewhat simpler - it's possible for the language to be extended to make this not so much of a problem, but providing defaults arguably the best way, since we still have to try infer types in any other case.


cmd/tasks_run.go
```diff
@@ -0,0 +1,180 @@
...
+		"The environments environment variables JSON payload")
+	tasksRun.Flags().StringVarP(&lagoonYml, "lagoon-yml", "l", ".lagoon.yml",
+		"The .lagoon.yml file to read")
+	tasksRun.Flags().BoolVarP(&runPreRollout, "pre-rollout", "", false,
+		"Will run pre-rollout tasks if true")
+	tasksRun.Flags().BoolVarP(&runPostRollout, "post-rollout", "", false,
+		"Will run post-rollout tasks if true")
+	tasksRun.Flags().StringVarP(&namespace, "namespace", "n", "",
+		"The environments environment variables JSON payload")
+	tasksRun.Flags().BoolVarP(&outOfClusterConfig, "out-of-cluster", "", false,
```
shreddedbacon, line 177: Could this be inverted logic so no flag is required?

We can update the way builds start in the future to create a `KUBECONFIG` instead of doing it the way we do currently (see: https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy.sh#L54-L59)

shreddedbacon: Have it check `KUBECONFIG`, if not found, fall back to checking the deployer token, if that isn't found then exit with error?

cmd/tasks_run.go
```diff
@@ -0,0 +1,180 @@
...
+	task.Service = incoming.Service
+	task.Shell = incoming.Shell
+	task.Container = incoming.Container
+	err := lagoon.ExecuteTaskInEnvironment(task)
+	return err
+}
+
+func init() {
+	configCmd.AddCommand(tasksRun)
+	tasksRun.Flags().StringVarP(&projectVariables, "project-variables", "v", "",
```
shreddedbacon, line 165: These flags are on `rootCmd` now

cmd/tasks_run.go
```diff
@@ -0,0 +1,180 @@
...
+		if runPostRollout {
+			fmt.Println("Executing Post-rollout Tasks")
+			err2, done := iterateTasks(lagoonConditionalEvaluationEnvironment, unwindTaskRun(lYAML.Tasks.Postrollout))
+			if done {
+				return err2
+			}
+			fmt.Println("Post-rollout Tasks Complete")
+		} else {
+			fmt.Println("Skipping post-rollout tasks")
+		}
```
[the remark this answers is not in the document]
bomoko: yep

cmd/tasks_run.go
```diff
@@ -0,0 +1,180 @@
...
+
+func init() {
+	configCmd.AddCommand(tasksRun)
+	tasksRun.Flags().StringVarP(&projectVariables, "project-variables", "v", "",
+		"The projects environment variables JSON payload")
+	tasksRun.Flags().StringVarP(&environmentVariables, "environment-variables", "V", "",
+		"The environments environment variables JSON payload")
+	tasksRun.Flags().StringVarP(&lagoonYml, "lagoon-yml", "l", ".lagoon.yml",
+		"The .lagoon.yml file to read")
+	tasksRun.Flags().BoolVarP(&runPreRollout, "pre-rollout", "", false,
```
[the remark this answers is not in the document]
bomoko: Sweet!

shreddedbacon:
I had a chat with @tobybellwood in standup about this after he showed example of it being used in a `.lagoon.yml` and have some thoughts:

We currently don't expose any build pod environment variables directly to the user, but we do convert some into `LAGOON_X` variables that get added to the config map if they are classed as `runtime|global`

What I'm thinking we should do here is instead of making all build pod environment variables available as comparison choices, we try and replicate what we do in builds (we will need to do this eventually anyway)

If we look at the current build script, we do this https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh#L746 a few times, where we populate a `values.env` file which gets loaded into the configmap alongside ones from the environment variables that are `runtime|global` https://github.com/uselagoon/lagoon/blob/main/images/kubectl-build-deploy-dind/build-deploy-docker-compose.sh#L1032-L1063

What I'm thinking here is that where we're crafting some existing values (https://github.com/uselagoon/build-deploy-tool/blob/main/cmd/helpers_values.go#L25-L38) we also craft a new instance of `[]lagoon.EnvironmentVariable` that we load up with these that we would be loading into the `values.env` file, and set their scope to be `runtime`. Then we merge these on top of the final generated variables https://github.com/uselagoon/build-deploy-tool/blob/main/cmd/helpers_values.go#L109 and this is then provided to anything that needs to consume them (eventually whatever we build to craft the configmap) including your `where` conditional checker

bomoko:
@shreddedbacon - I think the most recent commits bring all that in now - thanks for all your work on that.

cmd/tasks_run.go
```diff
@@ -0,0 +1,242 @@
...
+				return err, true
+			}
+		} else {
+			fmt.Printf("Conditional '%v' for task: \n '%v' \n evaluated to false, skipping\n", task.When, task.Command)
+		}
+	}
+	return nil, false
+}
+
+func getEnvironmentVariablesForConditionalEvaluation(pullWiderEnvironment bool) (tasklib.TaskEnvironment, error) {
```
shreddedbacon, line 157: Is this used anymore? Can we trim it and the associated tests?

shreddedbacon (commented):
One thing I'm not sure about with a comment in the review.

For test coverage, It would also be good to have a test that actually consumes an example `.lagoon.yml` and some predefined environment variable overrides like other cmd tests do. 
Added benefits of this is that there are some examples of what people can do in these example files

There are a few things that I think we will need to refine or refactor later on around collection of variables, but this impacts all the existing cmds too so not a blocker here.

cmd/tasks_run.go
bomoko: Good catch. Removed :)

bomoko:
> For test coverage, It would also be good to have a test that actually consumes an example `.lagoon.yml` and some predefined environment variable overrides like other cmd tests do. Added benefits of this is that there are some examples of what people can do in these example files

Yeah, we can definitely set that up - I've been trying to drive unit tests where possible, rather than full system (like) tests. But it's a great idea.

PR #71, opened 2022-07-25
feat: support ingress class if defined

### LINKED ISSUE #41: support additional IngressClass in ingresses
As an advanced user of Lagoon, managing a dedicated remote cluster, I may wish to install an additional Ingress. the build-deploy-tool should allow me to specify which IngressClass a non-default ingress be mapped to.

### LINKED ISSUE #70: Add IngressClass (and possible required annotations) to ingress creation process
As of Kubernetes 1.19 - both ingress-nginx and traefik require the presence of an IngressClass resource on created Ingresses if they are required to operate on a non-default Ingress.

Lagoon doesn't currently support this for workloads - to make this happen, we need three things - a feature flag (that can be set at the controller or proj/env level), a resource created in the ingress and any relevant Annotations (e.g. certmanager).  The first two are pretty straightforward - but we would have to find a way to define the annotation that needs setting as they may differ if people don't use certmanager?

shreddedbacon:
This feature adds the ability to set the class for ingress. It assumes that cert manager is used when setting the class for http solvers in the ingress annotations.

To utilise this feature, there are a number of ways, and some options are adjustable by developers/users of Lagoon environments, while others are reserved for platform operators.

The order in which some ways are defined are outlined below, some ways have precedence over others. This list is the order, with the one that takes precedence the most being at the bottom.

* `LAGOON_FEATURE_FLAG_DEFAULT_INGRESS_CLASS` (populated as the class name, ie `nginx`. platform operator only defined in `remote-controller`)
* `LAGOON_FEATURE_FLAG_INGRESS_CLASS`  (populated as the class name, ie `nginx`. defined in either project or environment scoped variables in the API)
* `LAGOON_FEATURE_FLAG_FORCE_INGRESS_CLASS` (populated as the class name, ie `nginx`. platform operator only defined in `remote-controller`)
* .lagoon.yml routes
	```
	environments:
  		main:
    		routes:
      			- node:
          			- example.com:
             			ingressClass: "custom-ingress"
	```
* `LAGOON_ROUTES_JSON` alpha feature way to inject or modify routes above .lagoon.yml

closes #70 
closes #41


tobybellwood:
I think here we need to be careful to make sure our terminology handles `defaultIngressClass` (i.e that defined in the cluster, by the ingress set as default), from any additional `ingressClass` (i.e. that defined in a given Ingress).

shreddedbacon:
> I think here we need to be careful to make sure our terminology handles `defaultIngressClass` (i.e that defined in the cluster, by the ingress set as default), from any additional `ingressClass` (i.e. that defined in a given Ingress).

This is still WIP, but if the default ingress class is not set (IE not defined in the remote-controller) then none is injected at all, like the value/annotation are not set. If a user defined the ingress class in the `.lagoon.yml` or LAGOON_ROUTES_JSON, then it will set this as the ingress class for only those defined.

If the default is set, but a user still defined a specific ingress class on 1 specific ingress, it will use the default on all but that 1 specific ingress.

shreddedbacon:
@tobybellwood are you alright with the order of precedence in the variables/overrides that this sets (in the main post). Or do you have a different preference?

It might seem weird that the `FORCE` flag is overridden by the `.lagoon.yml` and the routes JSON. But I figured that if someone has defined it in the `.lagoon.yml` or routes JSON it has been done for a reason. And leaving the routes JSON as the final override means it is possible to change what is defined in the `.lagoon.yml` if required.

tobybellwood:
I'm torn on this - I do want FORCE to mean force, and allowing JSON to override that may set a precedent?

shreddedbacon:
The main problem is the feature flag reading is a generic function that checks the feature flag variables only, it doesn't know about the extra logic that may extend deeper into lagoon. Extending that function or making it return additional variables for another function, and then building the logic that ignores the `.lagoon.yml` if the FORCE flag is set just complicates things a bit more.

shreddedbacon:
This is also the first instance that I am aware of of a feature flag that *could* also modify the values in the `.lagoon.yml`.
I feel like maybe in this case, it is acceptable?

tobybellwood:
maybe we accept that some feature flags aren't suitable for FORCEing?

But yes, I'm happy to go with your recommendation!

shreddedbacon:
> maybe we accept that some feature flags aren't suitable for FORCEing?

I think this is one of them. Just on the chance that the force is set, there is absolutely no way to modify a specific route to use a different one any other way. For routes, that could be very problematic.

PR #74, opened 2022-08-01
fix: label values longer than 63 characters

shreddedbacon:
If label values are longer than 63 characters, generating templates should fail.


smlx:
The `apimachinery` package has a bunch of validation functions that might be useful. In this case you might be interested in https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsValidLabelValue ?

shreddedbacon:
Ah nice

PR #76, opened 2022-08-04
Retain order of services in docker-compose file

### LINKED ISSUE #75: Random `LAGOON_ROUTE` with multiple, auto-generate-only, services
# Describe the bug

When deploying an environment that only has autogenerate routes, the `LAGOON_ROUTE` is not always for the first service listed in `docker-compose.yml`.

## To Reproduce

- Have a `docker-compose.yml` with multiple routable services (e.g., nginx and varnish)
- Run `build-deploy-tool identify primary-ingress`
- See that the primary ingress will randomly return the nginx or varnish route

## Expected behavior

The `LAGOON_ROUTE` should always be for the first service in the `docker-compose.yml`, assuming no other primary route has been defined.

## Additional context

The [type for unmarshaled docker compose services is a map](https://github.com/uselagoon/build-deploy-tool/blob/main/internal/generator/buildvalues.go#L44), which is randomly ordered. Therefore, when [iterating over the services](https://github.com/uselagoon/build-deploy-tool/blob/4e775e66f703d6afd321dbebb4374a0442afd37d/internal/generator/ingress.go#L135) to create autogenerated routes, the order will be random. When no primary route is defined, the first auto-generated route is picked to be the primary route and cause the returned primary route to be random.

shreddedbacon:
Lagoon relies on the order of the services in the docker-compose file to build images, and if no routes are defined it uses the order in the docker-compose to determine the primary autogenerated ingress.

Thanks @rocketeerbkw for identifying the issue and crafting the test case

closes #75


Makefile
```diff
@@ -8,7 +8,7 @@ PKG=github.com/uselagoon/build-deploy-tool
 LDFLAGS=-w -s -X ${PKG}/cmd.bdtVersion=${VERSION} -X ${PKG}/cmd.shortCommit=${SHORTCOMMIT} -X ${PKG}/cmd.bdtBuild=${BUILD} -X "${PKG}/cmd.goVersion=${GO_VER}"
 
 test: fmt vet
-	go test ./...
+	go clean -testcache && go test -v ./...
```
rocketeerbkw, line 11: I only made this change to verify that the test was failing randomly and that I wasn't crazy. I don't actually know if it's a good idea to keep?

rocketeerbkw: Otherwise it LGTM

shreddedbacon: It is fine, it won't impact CI tests because they will all be fresh cache anyway. For local testing it is also fine, so I left it in

PR #81, opened 2022-08-14
Feature - Ability to override pre/post rollout tasks with env vars and lagoon.override.yml files

### LINKED ISSUE #73: Allow .lagoon.yml override/merge for pre and post deploy tasks
Proposal:

It would be useful to be able to centrally control and add _extra_ pre and post rollout tasks, over and above those explicitly defined in the `.lagoon.yml`.

The proposal is to allow merging of pre and post rollout tasks via two avenues.

1.  having something like a docker compose override file, a `.lagoon.override.yml` (for example) whose entries will be merged into/over a given `.lagoon.yml` file.
2. Having an environment variable (name still tbd) which will itself contain the contents of a lagoon.yml override. This will allow the specification of extra tasks to be done via the api, rather than needing to add a new file to the git repo.

Any additional pre/post rollout tasks defined here will be run as part of the deployment.

bomoko:
This PR introduces the ability to override pre and post rollout tasks with either a `lagoon.override.yml` (default name can be overridden by args) file or an environment variable holding a base64 encoded lagoon.yml file (by default named `LAGOON_YAML_OVERRIDE`, also overridable).

closes #73


cmd/root.go
```diff
@@ -103,6 +103,10 @@ func init() {
 
 	rootCmd.PersistentFlags().StringVarP(&lagoonYml, "lagoon-yml", "l", ".lagoon.yml",
 		"The .lagoon.yml file to read")
+	rootCmd.PersistentFlags().StringVarP(&lagoonYmlOverride, "lagoon-yml-override", "", ".lagoon.override.yml",
+		"The .lagoon.yml override file to read for merging values into target lagoon.yml")
+	rootCmd.PersistentFlags().StringVarP(&lagoonYmlEnvVar, "lagoon-yml-envvar", "", "LAGOON_YAML_OVERRIDE",
```
shreddedbacon, line 108: Do you think this would be something people would want to change?

bomoko: @shreddedbacon - yeah, I think it's _possible_ - happy to move it into the validate command only if you want (which is where I think it'll mostly be used, if at all).
I just wired it up like this to make it a little more flexible.

shreddedbacon: I'm just struggling to find why someone would want to use a different named envvar to populate?

bomoko: @shreddedbacon - I guess it's really less about the name than I wondered about making it fixed. I'll pull it out.
I think the override file name is probably useful though.

shreddedbacon: Yeah, I'm ok with filename override. If someone was doing something locally they're more likely to be using files than base64 encoding variables, so it makes more sense to keep the filename flag. But if they're going to the point of base64 encoding values, I think supporting just 1 envvar is perfectly fine.

bomoko: @shreddedbacon done :)

PR #87, opened 2022-08-18
Feature: Spot instance support

tobybellwood:
Replaces https://github.com/uselagoon/lagoon/pull/3246

Add support for spot instances.

New feature flags are added to enable/disable spot instance support, the `TYPES` are a comma separated list of supported template types, spot is configurable between production and development so each can have different services on spot instances.
```
LAGOON_FEATURE_FLAG(_DEFAULT|_FORCE)_SPOT_INSTANCE_DEVELOPMENT=enabled
LAGOON_FEATURE_FLAG(_DEFAULT|_FORCE)_SPOT_INSTANCE_DEVELOPMENT_TYPES=nginx,nginx-persistent,nginx-php,nginx-php-persistent
LAGOON_FEATURE_FLAG(_DEFAULT|_FORCE)_SPOT_INSTANCE_DEVELOPMENT_CRONJOB_TYPES=cli,cli-persistent
LAGOON_FEATURE_FLAG(_DEFAULT|_FORCE)_SPOT_INSTANCE_PRODUCTION=enabled
LAGOON_FEATURE_FLAG(_DEFAULT|_FORCE)_SPOT_INSTANCE_PRODUCTION_TYPES=nginx,nginx-persistent,nginx-php,nginx-php-persistent
LAGOON_FEATURE_FLAG(_DEFAULT|_FORCE)_SPOT_INSTANCE_PRODUCTION_CRONJOB_TYPES=cli,cli-persistent
```

By default if spot is enabled for a specific service type, only the preference tolerations are added. If you want to force the pods to get a nodeselector for spot, you can add `:force` to the end of the type you want to force onto spot nodes.

Example: the bellow would allow all `nginx` types to go onto spot, but only `nginx` and `nginx-php` will get the nodeselector added to them.
```
LAGOON_FEATURE_FLAG(_DEFAULT|_FORCE)_SPOT_INSTANCE_DEVELOPMENT_TYPES=nginx:force,nginx-persistent,nginx-php:force,nginx-php-persistent
```

An example yaml of an nginx-php-persistent with the changes to the deployment that this PR does adding the `lagoon.sh/spot: true` label, and the tolerations for spot.
```
spec:
  replicas: 2
  selector:
    matchLabels:
      app.kubernetes.io/instance: nginx
      app.kubernetes.io/name: nginx-php-persistent
  template:
    metadata:
      labels:
        lagoon.sh/spot: 'true'
    spec:
      tolerations:
        - key: lagoon.sh/spot
          operator: Exists
          effect: NoSchedule
        - key: lagoon.sh/spot
          operator: Exists
          effect: PreferNoSchedule
```

Another feature request was to allow the replicas to change for production environments. This first implementation has hardcoded the following, as not all service types could support multiple replicas. This could also be turned into a separate feature flag variable.
```
SPOT_REPLICA_TYPES=nginx,nginx-persistent,nginx-php,nginx-php-persistent
```

Still to do support for:
- [ ] Cronjobs
- [ ] Backups

# Closing issues

partially closes uselagoon/lagoon#3219


shreddedbacon:
This is running on some test infrastructure that supports spot instances. One thing to note is these tolerations do not force pods onto spot instances, and won't force spot instance nodes to be created as they have a preference, not a requirement, for spot.

Schnitzel:
For any cluster that uses cluster autoscaler with pools for normal nodes and pools for spot nodes this code will never create a spot instance for these pods. If we want the cluster to profit from spot instances we need to have a nodeselector on the spot instances.

seanhamlin:
We might need a Affinity declared in the pods with `preferredDuringSchedulingIgnoredDuringExecution` (if not forced) to prefer Spot. See https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity

PR #88, opened 2022-08-18
Feature: Reduce build log verbosity

tobybellwood:
migrated from uselagoon/lagoon#3243

# Checklist

- [ ] Affected Issues have been mentioned in the Closing issues section
- [ ] Documentation has been written/updated
- [ ] PR title is ready for inclusion in changelog

The main goal is to reduce the entirety of the `set +x/-x` usage and only display useful information. Previously there was too much debugging information which with all the `set +x/-x` usage would obscure a lot of the warnings. A lot of YAML is also dumped straight into stdout for no real reason beyond being able to see what is created, for most users this is unused or just causes bulk on build logs for no reason.

This PR introduces a new flag though which can be set in the controller or as an environment variable to expose the created YAML files if required:
* `LAGOON_FEATURE_FLAG_FORCE_BUILD_VERBOSITY=enabled` (controller)
* `LAGOON_FEATURE_FLAG_BUILD_VERBOSITY=enabled` (environment variable)
* `LAGOON_FEATURE_FLAG_DEFAULT_BUILD_VERBOSITY=enabled` (controller)

An example build log is available here: https://gist.github.com/shreddedbacon/2ebb0f5cbc0ab524813959de1436ea2c


shreddedbacon:
Closing, we can re-visit verbosity under different work in the future, but #334 did some of what this does.

PR #95, opened 2022-08-24
refactor: some booleans can be strings in lagoon.yml

shreddedbacon:
Some cases where users have been able to define strings in the past is in the `routes.autogenerated.enabled` field in lagoon.yml. This field can be a string or a bool, so we need to cater for that in the tool.

Previous versions of this tool silently failed and hid this error until recently when we returned these errors.


tobybellwood:
confirmed working now

Issue #56, opened 2022-07-07
Fallback to db-single flag with default disable


shreddedbacon:
I think we should disable the fallback to `xdb-single` by default, and have a flag to enable the fallback feature in
* .lagoon.yml
  * `allowDatabaseFallback: true|false` ?
* docker-compose service label
  * `lagoon.databasefallback: true|false`
* lagoon api environment variable
  * `LAGOON_FEATURE_FLAG_DATABASE_FALLBACK: enabled|disabled|true|false`

This would mean that builds will FAIL if there is no dbaas-provider present to take the workload, which seems desirable as it would prevent builds from creating db pods and then trying to figure out what went wrong and fixing the environment to use a dbaas provider.

shreddedbacon:
With work I'm doing on the dbaas-provider configuration in this tool, and looking at past experiences in platform with builds that fail to provision against a dbaas for any number of reasons (requested environment type doesn't have a provider, the dbaas-operator is not responding/not installed), a build failure would have been much better and caused less issues for fixing it after the fact.

I know that we do have some users that WANT db-single pods, and they've already accounted for these in their docker-compose files or lagoon.yml as overrides, this behaviour wouldn't change.

seanhamlin:
Perhaps fail should be the default, and only if there is some positive action by the user (e.g. what you have above) then it should continue with a pod. Agree, this for larger clusters is a pain.

shreddedbacon:
Yep, I'm proposing failure is the default

shreddedbacon:
@tobybellwood would be good to get some thoughts on this, so I can potentially roll this into #115

shreddedbacon:
bump @tobybellwood

tobybellwood:
I agree that a fallback default is a bad thing:tm: , and should only really be enabled at a cluster level (to account for a missing dbaas). If a project, or branch, wants a single - they can define it in their docker-compose - I don't think we need a complex triple override system?

Issue #27, opened 2022-05-17
Migrate existing KBDD functionality from Bash to golang


tobybellwood:
The existing build stages that are currently in Bash, that need migrating across to golang here:

- [x] Custom route generation (https://github.com/uselagoon/lagoon/pull/3133)
- [x] Autogenerated route generation (https://github.com/uselagoon/lagoon/pull/3141)

Additional functionlity to be added to the build-tool

- [ ] AlternativeNames to add to custom routes (#25)

shreddedbacon:
https://github.com/uselagoon/lagoon/issues/3052 is another thing we should build into here, rather than try back port into the bash

PR #115, opened 2022-09-12
Change: use build-tool for generating dbaas consumers in builds

shreddedbacon:
Use build-deploy-tool to handle the creation of dbaas templates. This replaces the existing BASH process.

At the moment this is replacement of existing functionality, but I think it should also force failure of build on failure to provision dbaas https://github.com/uselagoon/build-deploy-tool/issues/56, but this could be a follow up PR

Sub task of https://github.com/uselagoon/build-deploy-tool/issues/27


internal/generator/services.go
```diff
@@ -35,6 +35,16 @@ var supportedAutogeneratedTypes = []string{
...
 }
 
+// these are lagoon types that support autogenerated routes
+var supportedDBTypes = []string{
+	"mariadb",
+	"mariadb-dbaas",
+	"postgres",
+	"postgres-dbaas",
+	"mongodb",
+	"mongodb-dbaas",
```
tobybellwood, line 45: There may also be a "mongo" type as per https://github.com/uselagoon/build-deploy-tool/blob/main/legacy/build-deploy-docker-compose.sh#L558

and this is how we built the internal-services-test (https://github.com/uselagoon/internal-services-test/blob/main/docker-compose.yml#L81)

shreddedbacon: :hidethepain:

tobybellwood: (confirmed that `lagoon.type: mongo` is the more commonly used option, but it should also work for mongodb too

tobybellwood (approved):
This is looking good from my side. Tested with all DBaaS providers, in new and update scenarios. The correction of the environment labels is nice, and the addition of the annotations to the postgres consumer both work perfectly. Additionally it clears up a situation where the underlying image for the DBaaS-ed service was being copied into Harbor unnecessarily

Great piece of work. A++

PR #119, opened 2022-09-19
try connect to docker-host a few times

shreddedbacon:
After the introduction of network policies in `lagoon-remote` on the docker-host, there is a race condition where the network policy hasn't applied fully before the pod has started. This results in the docker-host check failing.

To fix this, the build should try connect to the docker-host a few times before failing.


smlx (approved):
LGTM and works to fix the issue we saw with network policies taking a few seconds to take effect.

PR #125, opened 2022-10-04
Change: use build-deploy-tool for backup schedule generation

shreddedbacon:
This replaces the bash script used for handling backup schedules with the build-deploy-tool command.

Also bundled into this is the creation of `PreBackupPods` rather than via the helm templates. This brings all of the backup related generation to one command.

It is currently wrapped in a feature flag wrapper to use, you can set this on an project, or environment, or define it in the `remote-controller` with `DEFAULT|FORCE`. 
```
LAGOON_FEATURE_FLAG_K8UP_V2=enabled
```

The default is to fall back to the previous behavior of checking if k8up v1.x is installed. Eventually v1.x support will go away.


shreddedbacon:
I re-deployed one of our test environments in our test cluster that has k8up v1.x and k8up v2.x installed.

The build created all the new k8up.io resources and removed the older backup.appuio.ch resources from the namespace. Backups have since completed in this environment and appear in the Lagoon UI and are retrievable.

shreddedbacon:
This changes quite a bit about how the templates are generated, I've tried to cover off as many of the potential cases with tests that generate the schedules and prebackuppods that were previously using helm.

Would be good to get some extra :eyes: on this to see that it isn't going to change much else. I've done basic test coverage for the things I could think about, but happy to add more if I'm missing some that I couldn't think of

internal/templating/backups/template_prebackuppod.go
```diff
@@ -0,0 +1,426 @@
...
+        valueFrom:
+          configMapKeyRef:
+            key: {{ .Name | VarFix }}_PASSWORD
+            name: lagoon-env
+      - name: BACKUP_DB_DATABASE
+        valueFrom:
+          configMapKeyRef:
+            key: {{ .Name | VarFix }}_DATABASE
+            name: lagoon-env
+      image: imagecache.amazeeio.cloud/amazeeio/alpine-mysql-client
```
tobybellwood, line 293: podspec is missing `BACKUP_DB_READREPLICAS` var introduced in https://github.com/uselagoon/build-deploy-tool/pull/184

shreddedbacon: https://github.com/uselagoon/build-deploy-tool/pull/125/files#diff-d2bb5cf6b1385ed8e8b7a4b1e08c7333394c71ca58d53895ef410f7b909f10aeR90 because read replicas are optional in the dbaas, the switch for this happens in code

PR #126, opened 2022-10-05
Change: use pg_dump for postgres-single

shreddedbacon:
`mariadb-dbaas` and `mariadb-single` both use mysqldump to produce backups

`postgres-dbaas` does pg_dump, but `postgres-single` backs up the data volume, which doesn't seem very useful in terms of being able to easily move to another database.

This brings `postgres-single` in line to use pg_dump like `postgres-dbaas`.

Unfortunately, this would mean that backups from after users deploy after this change would have a different type of backup file to restore with. But I hope this would be a welcome change for people to be able to easily restore from.


shreddedbacon:
@tobybellwood do you see any concerns about doing this?

tobybellwood (approved):
Makes sense to me!

PR #127, opened 2022-10-12
Feature: add OpenSearch template

tobybellwood:
Now that Lagoon (shortly) supports OpenSearch images, this PR adds an OpenSearch lagoon.type for use with it. It also updates the Elasticsearch init container to one that is better supported in GKE clusters

replaces https://github.com/uselagoon/lagoon/pull/3126


shreddedbacon (changes requested):
Can you update this to support spot instances like the other templates: https://github.com/uselagoon/build-deploy-tool/pull/87

shreddedbacon (approved):

PR #137, opened 2022-10-31
Feature: Add LAGOON_BUILD_NAME to BUILD_ARGS

bomoko:
https://github.com/uselagoon/lagoon/issues/2246 raised the question of adding randomization in a build arg in order to break caching. This PR (at the suggestion of @tobybellwood ) injects the build pod's hostname (which is unique) as the variable `LAGOON_BUILD_NAME`.


tobybellwood:
And then all someone needs do is call
```
ARG LAGOON_BUILD_NAME
RUN echo $LAGOON_BUILD_NAME > /lagoon/lagoon-build
```
at the point they want to break caching

shreddedbacon (approved):
This seems like a fairly low impact change

PR #140, opened 2022-11-03
chore: support consuming deployer-token from service account

shreddedbacon:
In kubernetes 1.24, service accounts no longer get automatically generated token secrets. It is possible to generate long lived tokens, but using the TokenRequest API is the recommended way forward.

The remote-controller will now use the serviceaccount on all pods started by it, which uses volume projection to mount the token into the pod automatically.

The default will be to try the existing token location, but if one has been provided the service account, it will use this this.


tobybellwood (approved):
I'm happy with this for now, given then need to support <1.21

Once we have a 1.21/1.22 minimum, we can remove the lagoon/deployer/token completely, and just rely on the bound service account token at the new location.

PR #141, opened 2022-11-07
Feature: unidle namespace for pre-rollout tasks

bomoko:
This PR introduces namespace unidling for pre-rollout tasks.

This will ensure that any services that should be available when we're trying to run pre-rollout tasks (assuming the deployments exist) should be running.


shreddedbacon:
Sorry, I clicked the wrong button and merged in before asking if you were ready to go for this.

shreddedbacon:
> @shreddedbacon so this should work now, but I'm wondering if we want or need to wait for the pods to actually be ready?

Say someone runs a task in their CLI pod that needs also communicate with other pods in the environment. If you don't wait for them to start, then the task will fail if the communication to those other pods fails. You might get lucky by not waiting, but it isn't a guarantee that they will be started in time for the task to run.

bomoko:
yep, that was precisely my worry!

cmd/tasks_run.go
```diff
@@ -42,7 +43,29 @@ var tasksPreRun = &cobra.Command{
...
+		// We actually want to unidle the namespace before running pre-rollout tasks,
+		// so we wrap the usual task runner before calling it
+		unidleThenRun := func(namespace string, incoming lagoon.Task) error {
+			err := lagoon.UnidleNamespace(context.TODO(), namespace, namespaceUnidleWaitInSeconds, namespaceUnidleWaitInSeconds)
+			if err != nil {
+				switch {
+				case errors.Is(err, lagoon.NamespaceUnidlingTimeoutError):
+					fmt.Println("Namespace unidling is taking longer than expected - this might affect pre-rollout tasks that rely on multiple services")
+				default:
+					return err
```
shreddedbacon, line 56: I'm a bit nervous about this one, basically if there was an error talking to kubernetes here, that will fail the pre-rollout tasks and return whatever error is presented to the user?

If that is the case, could we at least wrap the error in something that makes it clear that there was something unrelated to their codebase that caused the unidling to fail?

shreddedbacon: This would also block the task from running?

shreddedbacon (approved):

PR #142, opened 2022-11-10
fix: cronjob templates

shreddedbacon:
There is an issue with cronjob templates where filling out the affinity, nodeselector, or tolerations will fail. This corrects the templating issues.


tobybellwood (approved):
Smashing the Approve Button

PR #143, opened 2022-11-10
add build-time DOCKER_BUILDKIT var

tobybellwood:
This PR adds support for `build` scoped env var, `DOCKER_BUILDKIT` to be set - usually to "1", undertaking the image build using the BuildKit enhancements.

It's currently set to off (0) by default but will be set to 1 on any "truthy" setting for the variable as per [strconv.ParseBool](https://pkg.go.dev/strconv#ParseBool)

I created a simple example here that uses buildkit to create a mount - https://github.com/amazeeio-demos/demo-python-simple-flask/tree/buildkit


tobybellwood:
:ninja: - likely of interest to @back-2-95 and @bryangruneberg - will ping you once it's generally available.

legacy/build-deploy-docker-compose.sh
```diff
@@ -698,6 +698,28 @@ if [[ "$BUILD_TYPE" == "pullrequest"  ||  "$BUILD_TYPE" == "branch" ]]; then
...
+      set +x # reduce noise in build logs
+      # Decide whether to use BuildKit for Docker builds - disabled by default.
+      DOCKER_BUILDKIT=0
+
+      if [ ! -z "$LAGOON_PROJECT_VARIABLES" ]; then
+        DOCKER_BUILDKIT=($(echo $LAGOON_PROJECT_VARIABLES | jq -r '.[] | select(.scope == "build") | select(.name == "DOCKER_BUILDKIT") | "\(.value)"'))
+      fi
+      if [ ! -z "$LAGOON_ENVIRONMENT_VARIABLES" ]; then
+        DOCKER_BUILDKIT=($(echo $LAGOON_ENVIRONMENT_VARIABLES | jq -r '.[] | select(.scope == "build") | select(.name == "DOCKER_BUILDKIT") | "\(.value)"'))
+      fi
```
shreddedbacon, line 710: [You need to do this sort of pattern](https://github.com/uselagoon/build-deploy-tool/blob/v0.19.4/legacy/build-deploy-docker-compose.sh#L269-L277). Otherwise it will almost always be 0 due to how the two if statements evaluate.

tobybellwood: I'll relook at it - technically it does work, because the environment not being set doesn't evaluate as truthy, so gets reset to 0 anyway.

shreddedbacon: See these two examples https://gist.github.com/shreddedbacon/f418d84c7e3feff47bdc241308a503ea

shreddedbacon: I just added a third example using the TEMP_X pattern to show you the difference between

tobybellwood: ugh :facepalm: - of course, if there's no ENV vars set, it gets skipped, hence my pass!

shreddedbacon (approved):

back-2-95:
How it goes from here now that it's merged?

tobybellwood:
the next release of Lagoon (in the next couple of weeks) will bring this into production - however, we may deploy the build image earlier than that - will let you know if we do.

back-2-95:
When I see this on dashboard `Lagoon core-v2.10.0+46e4aa51` < to what commit this refers to?

tobybellwood:
> When I see this on dashboard `Lagoon core-v2.10.0+46e4aa51` < to what commit this refers to?

annoyingly, thats the UI version (https://github.com/uselagoon/lagoon-ui/releases/tag/core-v2.10.0) we're working on a more advanced panel to show the Lagoon versions etc.

the builds themselves show the current version in the logs (under Logs Begin):

Our test cluster (where this is running) is here, but our prod clusters will be behind it until we re-release a build image
```
build-deploy-tool v0.19.4+ccca1a61 (built: 2022-11-10T06:42:59+0000 / go 1.17)
```

PR #148, opened 2022-11-27
Feature: dynamic secret loading

bomoko:
This PR introduces the notion of a dynamically loaded secret. It allows a new way of providing volume mounted secrets to pods by searching through the target namespace at build-deploy time for any secrets labelled with `lagoon.sh/dynamic-secret`
Once found, this secret will then be volume mounted at `/var/run/secrets/lagoon/dynamic`.

The use case for this is straightforward - it allows, for example, services that live in lagoon-remote the opportunity to add cluster specific secrets into pods without needing to communicate these data back to the lagoon core.
The specific use case this is being developed for is to allow the `insights-remote` system to inject a token that can be used inside the cluster specific pods to authorize communication back to the cluster local insights-remote instance.


shreddedbacon (approved):
Looks reasonable to me

PR #152, opened 2022-12-05
feature: add primaryingress label to primary ingress

### LINKED ISSUE #149: Primary ingress domains shoud be labled
To make monitoring sites easier and give the possibility to filter for primary domains, we should start adding a label for those ingresses.

Proposal
```
lagoon.sh/primaryingress=true
```

shreddedbacon:
Adds the `lagoon.sh/primaryingress` label to ingress that are detected as being the primary.

closes #149


shreddedbacon:
ugh, what happened with the diff for this

PR #153, opened 2022-12-06
chore: add primaryingress label to primary ingress

### LINKED ISSUE #149: Primary ingress domains shoud be labled

shreddedbacon:
closes #149


tobybellwood (approved):
this makes sense. It'll tie in nicely with an improvement to the route handling

internal/templating/ingress/test-resources/result-custom-ingress1.yaml
```diff
@@ -26,6 +26,7 @@ metadata:
     lagoon.sh/buildType: branch
     lagoon.sh/environment: environment-with-really-really-reall-3fdb
     lagoon.sh/environmentType: production
+    lagoon.sh/primaryingress: "true"
```
smlx, line 29: This adds a third casing type to the labels: all lowercase. Currently the other labels are a mix of kebab-case and camelCase. I wonder if it should use one of the other standards?

shreddedbacon: Yeah, good spot. Missed this.

shreddedbacon: Most of the labels are camelcase, only one or two are kebab case.

Not sure which one to standardise on though at this point. But camelcase has more than the other

tobybellwood: camelCase makes more sense IMO - I'd only suggest kebab if there's a logical separation between words?

PR #156, opened 2022-12-07
Increases verbosity of logging for task service and container

### LINKED ISSUE #155: Change output of pre and post rollout task logs
If a user tries to run a pre or post rollout task in an `nginx-php-persistent` or any other `x-persistent` service, it will fail because there are 2 or more containers in the pod (with rootless workloads, there is the init container, with php there are three containers)

The [docs ask a user to provide the `service`](https://docs.lagoon.sh/using-lagoon-the-basics/lagoon-yml/#post-rollout-tasks-post_rolloutirun) to run the task in, but in no place do we mention the container to also use.

[We call out this problem with SSH](https://docs.lagoon.sh/using-lagoon-advanced/ssh/#podservice-container-definition), so we should support this option for tasks too

Example: requires the user to specify `nginx`, `php`, or `fix-storage-permissions` (the last one being the init container for rootless)
```
Executing Post-rollout Tasks
Executing task 'Deletes all cached pages':'rm -rf storage/cache/blitz'
Going to exec into  nginx-5fd488b7dc-gbj52
*** Task 'Deletes all cached pages' failed - STDOUT and STDERR follows ***
Post-rollout Tasks Failed with the following error:  Error returned: a container name must be specified for pod nginx-5fd488b7dc-gbj52, choose one of: [fix-storage-permissions nginx php]
```

bomoko:
closes #155

### DIFF
--- a/internal/lagoon/tasks.go
+++ b/internal/lagoon/tasks.go
@@ -234,7 +234,7 @@ func ExecTaskInPod(
 		}
 	}
 	if debug {
-		fmt.Printf("Executing task '%v' in pod %v \n", task.Name, pod.Namespace)
+		fmt.Printf("Executing task '%v' in pod %v \n", task.Name, pod.Name)
 	}
 
 	req := clientset.CoreV1().RESTClient().Post().
--- a/internal/lagoon/tasks.go
+++ b/internal/lagoon/tasks.go
@@ -234,7 +234,11 @@ func ExecTaskInPod(
 		}
 	}
 	if debug {
-		fmt.Printf("Executing task '%v' in pod %v \n", task.Name, pod.Name)
+		podName := pod.Name
+		if task.Container != "" {
+			podName = fmt.Sprintf("%v/%v", podName, task.Container)
+		}
+		fmt.Printf("Executing task '%v' in pod %v \n", task.Name, podName)
 	}
 
 	req := clientset.CoreV1().RESTClient().Post().
--- a/internal/lagoon/tasks.go
+++ b/internal/lagoon/tasks.go
@@ -111,13 +111,6 @@ func getConfig() (*rest.Config, error) {
 
 // ExecuteTaskInEnvironment .
 func ExecuteTaskInEnvironment(task Task) error {
-	if debug {
-		podContainerDetails := fmt.Sprintf("Service:%v", task.Service)
-		if task.Container != "" {
-			podContainerDetails = fmt.Sprintf("%v and container:%v", podContainerDetails, task.Container)
-		}
-		fmt.Printf("Preparing to execute task named '%v' with definition '%v' in %v \n", task.Name, task.Command, podContainerDetails)
-	}
 	command := make([]string, 0, 5)
 	if task.Shell != "" {
 		command = append(command, task.Shell)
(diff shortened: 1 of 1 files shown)


tobybellwood:
This looks much better!
```
Executing Post-rollout Tasks
Preparing to execute task named 'env variables' with definition 'env' in Service:nginx and container:php 
Going to exec into  nginx-79965775c7-hkr6c
*** Task STDOUT ***
```
and
```
Preparing to execute task named 'drush cr' with definition 'cd web && drush -y cr' in Service:cli 
Going to exec into  cli-84c6d9c6d6-gvp9f
*** Task STDERR ***
```

Although lengthy task definitions make it harder to read
```
Preparing to execute task named 'IF no Drupal installed, install Drupal' with definition 'cd web
if tables=$(drush sqlq "show tables like \"node\";") && [ -z "$tables" ]; then
    # no drupal installed, we install drupal from an db dump
    zcat ../drupal.sql.gz | $(drush sql-connect)
fi
# set the sitename to the current SHA and BRANCH
drush -y cset system.site name "LAGOON_GIT_SHA=$LAGOON_GIT_SHA LAGOON_GIT_BRANCH=$LAGOON_GIT_BRANCH"
' in Service:cli
```

Perhaps we can shuffle the order a little to put the name, service and container on one line and the task on the next line(s)?

internal/lagoon/tasks.go
```diff
@@ -112,7 +112,11 @@ func getConfig() (*rest.Config, error) {
 // ExecuteTaskInEnvironment .
 func ExecuteTaskInEnvironment(task Task) error {
 	if debug {
-		fmt.Printf("Executing task '%v':'%v'\n", task.Name, task.Command)
+		podContainerDetails := fmt.Sprintf("Service:%v", task.Service)
```
tobybellwood, line 115: maybe lower case this for consistency?

```suggestion
		podContainerDetails := fmt.Sprintf("service:%v", task.Service)
```

shreddedbacon:
Does it need to output the command at all??

shreddedbacon:
Also, 
```
Preparing to execute task named 'drush cr' with definition 'cd web && drush -y cr' in Service:cli 
Going to exec into  cli-84c6d9c6d6-gvp9f
```

Could be one log like this?
```
Preparing to execute task named 'drush cr' in pod cli-84c6d9c6d6-gvp9f
```

bomoko:
@shreddedbacon It's two logs because the first announces what's going to be done - that way if there's an error trying to resolve the pod and container etc. we know what we were doing at the time it errored out.

I can change the errors in the exec process to rather say "failed to XYZ while attempting to execute task named T1 in service S1 and container C1" if that's preferable?

bomoko:
Happy to remove the actual comment as well

shreddedbacon:
> @shreddedbacon It's two logs because the first announces what's going to be done - that way if there's an error trying to resolve the pod and container etc. we know what we were doing at the time it errored out.
> 
> I can change the errors in the exec process to rather say "failed to XYZ while attempting to execute task named T1 in service S1 and container C1" if that's preferable?

Yeah, I think a failure message instead of the `preparing` that way its one of 2 things displayed in the log.

Successful task would print 
```
Executing task 'x' in pod 'y'
```

Where a failed task would print
```
Failed to execute task 'x` due to reason 'y'
```

bomoko:
Okay, so I've had to push the task name into the ExecPod (now ExecTaskInPod) function in order to have the task name exist in the same frame as the actual exec code. It will now no longer write either the service or the container names, but rather the podname, if the pod name is matched.

internal/lagoon/tasks.go
```diff
@@ -235,16 +230,17 @@ func ExecPod(
 	}
 	if !foundRunningPod {
 		return "", "", &PodScalingError{
-			ErrorText: "Unable to find running Pod for namespace: " + namespace,
+			ErrorText: "Unable to find running Pod for namespace: " + task.Namespace,
 		}
 	}
 	if debug {
-		fmt.Println("Going to exec into ", pod.Name)
+		fmt.Printf("Executing task '%v' in pod %v \n", task.Name, pod.Namespace)
```
shreddedbacon, line 237: `pod.Namespace` or `pod.Name`?

shreddedbacon:
```

shreddedbacon (approved):

PR #160, opened 2022-12-09
Fail builds if service names are not DNS1053 compliant

### LINKED ISSUE #157: Lagoon doesn't support dots in service names
Docker Compose supports `[a-zA-Z0-9\._\-]` in service names (https://github.com/docker/compose/pull/1624) as per the compose spec `"^[a-zA-Z0-9._-]+$"` (https://github.com/compose-spec/compose-spec/blob/master/schema/compose-spec.json#LL23C9-L23C28)

Lagoon however, can't handle periods in service names in at least two places:
1) Our shyaml `hyaml get-value services.$COMPOSE_SERVICE.labels.lagoon\\.name` tries to interpret the `.` as dot notation (need to escape the . with a `)
2) Any ingresses would end up with a different  subdomain (replace . with -)

CGoodwin90:
closes #157


internal/lagoon/compose.go
```diff
@@ -57,6 +59,9 @@ func UnmarshalLagoonDockerComposeYAML(file string) ([]OriginalServiceOrder, erro
 		// extract the services only
 		if item.Key.(string) == "services" {
 			for idx, v := range item.Value.(goyaml.MapSlice) {
+				if err := utilvalidation.IsDNS1035Label(v.Key.(string)); err != nil {
+					return nil, errors.New("service name cannot contain '.'")
```
seanhamlin, line 63: I think this will catch errors other than fullstops, e.g. being longer than 63 characters, starting with a digit etc. Perhaps a more generic error message "Service name is invalid. See the docs <link>" or something.

shreddedbacon: I don't want hardcoded docs links in here, would get old real fast.

But generally clearer message referencing something that could exist in our docs that is searchable like "Read the documentation around service naming requirements" and then we get our documentation updated to mention this somewhere too

shreddedbacon:
The test failure is a my bad on the logic in the test code. Will show you how to fix it next week.

internal/lagoon/compose.go
CGoodwin90: Yep good call! I've updated the error message now, let me know if you want any changes

shreddedbacon: Looks good to me, we probably need a followup PR in the main lagoon docs to mention this requirement

internal/lagoon/compose_test.go
```diff
@@ -82,15 +82,15 @@ func TestUnmarshaDockerComposeYAML(t *testing.T) {
...
 			wantOrder: []OriginalServiceOrder{
 				{Index: 0, Name: "cli"},
 				{Index: 1, Name: "nginx"},
 				{Index: 2, Name: "php"},
 				{Index: 3, Name: "mariadb"},
 				{Index: 4, Name: "redis"},
 				{Index: 5, Name: "chrome"},
-				{Index: 6, Name: "wait_dependencies"},
+				{Index: 6, Name: "wait-dependencies"},
```
shreddedbacon, line 93: This one is going to be interesting. You had to change this in the test to fix the tests. But this is valid docker-compose and this particular test case is sanitized from a customer. So their builds will now fail.

internal/lagoon/compose.go
```diff
@@ -57,6 +59,9 @@ func UnmarshalLagoonDockerComposeYAML(file string) ([]OriginalServiceOrder, erro
 		// extract the services only
 		if item.Key.(string) == "services" {
 			for idx, v := range item.Value.(goyaml.MapSlice) {
+				if err := utilvalidation.IsDNS1035Label(v.Key.(string)); err != nil {
```
shreddedbacon, line 62: I think we should also then check if the `lagoon.type` label is not `none` here too. Because `lagoon.type=none` do not get deployed, we shouldn't fail builds for people using services like `wait_dependencies` for containers they may use for local development only that don't make it into lagoon.

shreddedbacon (changes requested):
Generally OK, but after reviewing the fix for `wait_dependencies` to `wait-dependencies` I think this could cause failures for people for containers they may use locally only, not in Lagoon. We only want to block them if they're turned into Lagoon services.

shreddedbacon (approved):

PR #162, opened 2022-12-13
feat: bump memory requests for node and php charts

smlx:
Empirical evidence suggests that the median memory usage for php and node is much closer to 100Mi than 10Mi.


tobybellwood (approved):
This makes a lot of sense - Looking at our mem req % values across clusters we can safely up this to 100Mi without impacting pod density.

PR #164, opened 2022-12-29
migrate all cronjobs to batch/v1 apiVersion

### LINKED ISSUE #129: CronJob API version update to batch/v1
https://github.com/uselagoon/build-deploy-tool/blob/a0a2b34c3e9fe08982709e8c4f8b0e07b05de684/legacy/helmcharts/cli/templates/cronjob.yaml#L3

The apiVersion for CronJobs in `legacy/helmcharts/cli/templates/cronjob.yaml` references the depreciated `batch/v1beta1` version, which is no longer supported in v1.22+. Requesting an update to switch to using `batch/v1` that is supported in active and maintained Kubernetes versions.

tobybellwood:
In Kubernetes 1.25, the batch/v1beta1 apiVersion for Cronjob has been deprecated, and batch/v1 should be used instead. There are no notable changes so this is a straight api upgrade.

Note that this will require a minimum Kubernetes version 1.21 to run, in line with Lagoon 2.12 release.

Closes #129


shreddedbacon (approved):

shreddedbacon:
This is good to go, but if merged now we need to make sure that all versions of released things note the 1.21+ requirement

tobybellwood:
updated all chart reqs in https://github.com/uselagoon/lagoon-charts/pull/526

shreddedbacon:
Once that is merged, then we can merge this one. I'd hate to release this before the minimum is released.

tobybellwood:
I've had to pin the version of openssh and its dependency openssh-client-common to 9.1_p1-r2 to get around a transient error in the Alpine package registry - this should resolve itself - I'll put a PR in to remove the pin once this has been merged.

PR #165, opened 2022-12-30
Error msg for builds without a Lagoon type

CGoodwin90:
The current error response for builds with no Lagoon type set is quite generic - e.g. `Rollout for ${SERVICE_NAME} failed, tried to gather some startup logs of the containers, but unfortunately there were none created, sorry.`

This provides a more verbose error message when no Lagoon type is set.

image

### DIFF
--- a/internal/generator/services.go
+++ b/internal/generator/services.go
@@ -90,9 +90,8 @@ func composeToServiceValues(
 		lagoonType = lagoon.CheckServiceLagoonLabel(composeServiceValues.Labels, "lagoon.type")
 	}
 	if lagoonType == "" {
-		return ServiceValues{}, fmt.Errorf("No Lagoon type has been set for service %s. If a Lagoon type is not required for this service, please set the Lagoon type to 'none'", composeService)
-	}
-	if lagoonType != "" {
+		return ServiceValues{}, fmt.Errorf("No lagoon.type has been set for service %s. If a Lagoon service is not required, please set the lagoon.type to 'none' for this service in docker-compose.yaml. See the Lagoon documentation for supported service types.", composeService)
+	} else {
 		// if the lagoontype is populated, even none is valid as there may be a servicetype override in an environment variable
 		autogenEnabled := true
 		autogenTLSAcmeEnabled := true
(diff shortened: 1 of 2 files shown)


internal/generator/services.go
```diff
@@ -89,6 +89,9 @@ func composeToServiceValues(
 	if composeServiceValues.Labels != nil {
 		lagoonType = lagoon.CheckServiceLagoonLabel(composeServiceValues.Labels, "lagoon.type")
 	}
+	if lagoonType == "" {
+		return ServiceValues{}, fmt.Errorf("No Lagoon type has been set for service %s. If a Lagoon type is not required for this service, please set the Lagoon type to 'none'", composeService)
```
tobybellwood, line 93:
```suggestion
		return ServiceValues{}, fmt.Errorf("No Lagoon type has been set for service %s. If a Lagoon service is not required, please set the lagoon.type to 'none' for this service in docker-compose.yaml", composeService)
```

shreddedbacon, line 93: I would change all the `Lagoon type` to be `lagoon.type` so that it matches the label name that should be used. and then putting a small note about reading the documentation for supported types. 

Something like this:
```
No lagoon.type has been set for service %s. If this service is not meant to be deployed in Lagoon, please set the lagoon.type to 'none'. See the Lagoon documentation for supported service types.
```

internal/generator/services.go
```diff
@@ -89,6 +89,9 @@ func composeToServiceValues(
...
+	}
 	if lagoonType != "" {
```
shreddedbacon, line 95: Probably change this to an else rather than doing the second if to check if it is not "".
```
} else {
```

shreddedbacon (changes requested):

internal/generator/services.go
shreddedbacon: Just saw Toby also commented on changes for this. So some sort of combo of the two would probably be good

CGoodwin90: Updated with a combo of this & Ben's

internal/generator/services.go
CGoodwin90: Updated now

internal/generator/services.go
CGoodwin90: Updated

shreddedbacon (approved):

PR #169, opened 2023-01-10
feat: flags to disable build sections

timclifford:
This adds support for build variables to enable or disable sections of a build from being run. There is a big caveat here that disabling these on builds that already have these features deployed will not remove them, it just prevents the steps that would generate them from being run. Existing defined resources will remain.

The following variables are available:
* `LAGOON_CUSTOM_ROUTES_DISABLED`
	* Setting this to true disables user defined ingress/routes being generated
* `LAGOON_AUTOGEN_ROUTES_DISABLED`
	* Setting this to true disables auto generated ingress/routes being generated
* `LAGOON_CRONJOBS_DISABLED`
	* Setting this to true disables user cronjobs being generated
* `LAGOON_BACKUPS_DISABLED`
	* Setting this to true disables backup configurations being generated


legacy/scripts/exec-backup-generation.sh
```diff
@@ -53,6 +53,16 @@ fi
 ##############################################
 
 
+set -x
+# Check if backups should be scheduled
+  if [ "$LAGOON_BACKUPS_DISABLED" = true ]; then
```
shreddedbacon, line 58: This will never be checked, you'll need to do something like this that checks the scope of the variable to see if it is a build variable: https://github.com/uselagoon/build-deploy-tool/blob/c9bed57808d0110b59f7d63b891e08a8e208c382/legacy/build-deploy-docker-compose.sh#L705-L713

timclifford: https://github.com/uselagoon/build-deploy-tool/pull/169/commits/137dfa02de52108899834445b7ab1dfe0f63b88a

timclifford:
These will be super useful 👍

shreddedbacon:
@tobybellwood just looping you in, @bomoko and I discussed merging these and think there is no harm in doing so, but good to get second :eyes: just to verify.

tobybellwood (approved):
yah - makes sense to me

PR #174, opened 2023-01-25
change: Convert elasticsearch/opensearch prebackuppods to dep commands

### LINKED ISSUE #172: Convert elasticsearch/opensearch prebackuppods to backup commands on the deployment
Currently our elasticsearch and opensearch templates both have prebackuppods configured, and they have the following spec command/extension: https://github.com/uselagoon/build-deploy-tool/blob/main/legacy/helmcharts/elasticsearch/templates/prebackuppod.yaml#L11-L12

There is no reason for these two templates to be configured with prebackuppods, so we should remove those templates, and convert them into the deployment template to match how solr does backup commands: https://github.com/uselagoon/build-deploy-tool/blob/main/legacy/helmcharts/solr/templates/deployment.yaml#L29-L30

We'll also need a way to clean up any elasticsearch/opensearch prebackuppods when the environments are deployed, so will need an additional check for the prebackuppods and then a followup to remove it if it exists, this could done at the same time as this bit happens: https://github.com/uselagoon/build-deploy-tool/blob/main/legacy/build-deploy-docker-compose.sh#L368-L370 just checking for `opensearch` or `elasticsearch` as the service type, then doing a kubectl get command on what the prebackuppod would be called in kubernetes at this point, if the object doesn't exist then it should not block a build

Something like this should be close to what should be checked at that step (might need to confirm the service_name, etc)
```
if [[ "${CAPABILITIES[@]}" =~ "backup.appuio.ch/v1alpha1/PreBackupPod" ]]; then
	if kubectl -n ${NAMESPACE} get prebackuppods.backup.appuio.ch "$SERVICE_NAME" &> /dev/null; then
		kubectl delete ...
	fi
fi
```

CGoodwin90:
Removed prebackuppod templates for Elasticsearch/Opensearch, and added backup commands to the Deployments for both.
Updated `build-deploy-docker-compose` to clean up any existing elasticsearch/opensearch prebackuppods when deployed.

closes #172


shreddedbacon (approved):
This looks good to me. The check for existing prebackupods looks good

PR #175, opened 2023-01-30
chore: add skip-triggers db backup command

### LINKED ISSUE #171: Database triggers exported twice in scheduled backups
# Describe the bug

Dumps of databases that have triggers contain the triggers twice. Once when creating the corresponding table and once after inserting data into the table. When such a dump is then imported onto a database it fails because it tries to create a trigger that already exists.

`ERROR 1359 (HY000) at line XXXXXX: Trigger already exists`

The root cause is that there's two dumps appended one after the other, the first containing the database structure and the other containing the actual data. For example, from [this template](https://github.com/uselagoon/build-deploy-tool/blob/main/legacy/helmcharts/mariadb-dbaas/templates/prebackuppod.yaml#L11-L30):

```
 backupCommand: >-
    /bin/sh -c "dump=$(mktemp)
    && mysqldump --max-allowed-packet=500M --events --routines --quick
    --add-locks --no-autocommit --single-transaction --no-create-db
    --no-data --no-tablespaces
    -h $BACKUP_DB_HOST
    -u $BACKUP_DB_USERNAME
    -p$BACKUP_DB_PASSWORD
    $BACKUP_DB_DATABASE
    > $dump
    && mysqldump --max-allowed-packet=500M --events --routines --quick
    --add-locks --no-autocommit --single-transaction --no-create-db
    --ignore-table=$BACKUP_DB_DATABASE.watchdog
    --no-create-info --no-tablespaces
    -h $BACKUP_DB_HOST
    -u $BACKUP_DB_USERNAME
    -p$BACKUP_DB_PASSWORD
    $BACKUP_DB_DATABASE
    >> $dump
```

In the second dump the `--skip-triggers` option should be added.

There may be other places in the code where this should be changed. I have had the issue occurring with scheduled backups i.e. found under the Backups tab in the Lagoon UI. Manual backups taken with Tasks do not seem to have this issue.

## To Reproduce

1. Create a database that has one or more triggers.
2. Configure scheduled backups.
3. Download a scheduled backup.
4. See that the backup file contains the trigger insert/update statements twice; once after the create table statement and again after the insert data statement.

## Expected behavior

A dump of a database with triggers should be imported without errors.

shreddedbacon:
closes #171


legacy/helmcharts/mariadb-dbaas/templates/prebackuppod.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     /bin/sh -c "dump=$(mktemp)
     && mysqldump --max-allowed-packet=500M --events --routines --quick
     --add-locks --no-autocommit --single-transaction --no-create-db
-    --no-data --no-tablespaces
+    --no-data --no-tablespaces --skip-triggers
```
krystalcode, line 15: The first command dumps the structure and the second command dumps the data. It might be better to include the triggers in the structure and not with the data? i.e. add `--skip-triggers` in the second command. Either would work I suppose.

shreddedbacon: Sure, happy to do that :)

tobybellwood (approved):
makes sense to me!

PR #181, opened 2023-02-03
revert openssh 9.1_p1-r2 pinning

tobybellwood:
in #164 we needed to pin the OpenSSH version to avoid an Alpine package resolution error - once this PR passes tests, we should merge it to remove that pin.

The conflict is caused by the upstream docker/20.10.22 image already having an older version of some OpenSSH libraries installed. These libraries have been updated in the Alpine package registry. Our attempt to install a newer version of the parent OpenSSH package, therefore, causes conflicts.

This PR, therefore, installs and upgrades all the OpenSSH libraries and packages in a single step.


shreddedbacon (approved):
yay :shipit:

PR #184, opened 2023-02-10
Change: Preference read replica for MariaDB & Postgres

### LINKED ISSUE #154: Preference the database read replica when creating a database dump if one is present
In the advent there is one or more read replicas for a given database cluster, we should prefer them for database backups. This will remove load from the writers on clusters where database backups are very frequent.

Unsure where the best place to do this, but perhaps passing in the read replicas as we as the writer

https://github.com/uselagoon/build-deploy-tool/blob/main/legacy/helmcharts/mariadb-dbaas/templates/prebackuppod.yaml#L58-L62

Then in the code, the pseudo logic I am thinking is

```shell
if (there is one or more read replicas) {
  pick a read replica at random from the replicas
}
else {
  use the writer
}
```

I think we make the assumption that the same credentials work on the writer as they do on the readers. This appears to be the case for both AWS RDS and Azure Flexible Server.

CGoodwin90:
Updated the MariaDB and Postgres prebackuppods to check for the presence of a read replica, and if one or more are availabile to utilize the first instance for backups.

closes #154


shreddedbacon (approved):

shreddedbacon:
This looks good to me, and I've seen this in use in the test infrastructure. 

Would be good to get a second set of :eyes: over it though just to be sure.

tobybellwood (approved):
approved as working in test - logic makes sense

PR #189, opened 2023-02-23
fix: set HSTS max-age to be 1 year by default

### LINKED ISSUE #188: Default HSTS max-age is too short
The [default HSTS `max-age`](https://github.com/uselagoon/build-deploy-tool/blob/df8b6d57492f29ba89514afb46df27978303714c/internal/lagoon/routes.go#L62) is `3153600` or ~2 months.

I think this is too short and should be changed to `31536000` (~1 year) because:

1. The minimum max-age for [preloading](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#preloading_strict_transport_security) is `31536000`
2. The [old `hsts` flag we used](https://github.com/uselagoon/build-deploy-tool/blob/df8b6d57492f29ba89514afb46df27978303714c/test-resources/identify-ingress/test16/lagoon.yml#L20) always had `31536000` as the default example

rocketeerbkw:
Sets HSTS `max-age` to 1 year. Removes deprecated `hsts` from test `.lagoon.yml` files.

Closes #188

### DIFF
--- a/internal/lagoon/routes.go
+++ b/internal/lagoon/routes.go
@@ -59,7 +59,7 @@ type Route struct {
 	Ingresses map[string]Ingress
 }
 
-var defaultHSTSMaxAge = 3153600
+var defaultHSTSMaxAge = 31536000
 
 // UnmarshalJSON implements json.Unmarshaler.
 func (r *Route) UnmarshalJSON(data []byte) error {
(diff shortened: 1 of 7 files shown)


shreddedbacon (approved):
Looks good

PR #192, opened 2023-03-07
fix: enforce only one instance of an in-pod cronjob running at once

smlx:
This change enforces that in-pod cronjobs have only one instance of the job running at any one time.

Specifically, this change fixes the issue where minutely cronjobs which take longer than a minute to run result in dozens of instances of the job running at the same time. In some cases this can cause Lagoon users to DoS themselves or use excessive CPU or memory resources.

Closes: https://github.com/uselagoon/lagoon-images/issues/172

### DIFF
--- a/internal/generator/services.go
+++ b/internal/generator/services.go
@@ -1,6 +1,7 @@
 package generator
 
 import (
+	"crypto/sha256"
 	"fmt"
 	"reflect"
 	"regexp"
@@ -8,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/alessio/shellescape"
 	composetypes "github.com/compose-spec/compose-go/types"
 	"github.com/uselagoon/build-deploy-tool/internal/helpers"
 	"github.com/uselagoon/build-deploy-tool/internal/lagoon"
@@ -479,7 +481,7 @@ func composeToServiceValues(
 			return nil, err
 		}
 		if !buildValues.CronjobsDisabled {
-			for _, cronjob := range buildValues.LagoonYAML.Environments[buildValues.Branch].Cronjobs {
+			for idx, cronjob := range buildValues.LagoonYAML.Environments[buildValues.Branch].Cronjobs {
 				// if this cronjob is meant for this service, add it
 				if cronjob.Service == composeService {
 					var err error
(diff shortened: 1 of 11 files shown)


smlx:
Not sure how to run this through the Lagoon test suite to confirm functionality. Any tips?

shreddedbacon:
> Not sure how to run this through the Lagoon test suite to confirm functionality. Any tips?

We do have a cronjob features test in the test suite, you could extend that to include a cronjob that would experience the issue that flock would solve. Not sure if this is the best way to do this though?

smlx:
What I mean is: how do I run this through the Lagoon test suite?

shreddedbacon:
I've updated this to use the service templates instead of bash. Will need to do some more testing, but seems to work well in the limited playing around I did.

internal/generator/services.go
```diff
@@ -492,6 +492,11 @@ func composeToServiceValues(
 					}
 					// if the cronjob is inpod, or the cronjob has an inpod flag override
 					if inpod || (cronjob.InPod != nil && *cronjob.InPod) {
+						cmd := cronjob.Command
+						// Lagoon enforces that only a single instance of a cronjob can run at any one time.
+						// https://man7.org/linux/man-pages/man1/flock.1.html
+						// https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion
+						cronjob.Command = fmt.Sprintf("flock -n /tmp/cron.lock.$(echo \"%d %s\" | sha256sum | awk '{ print $1 }') -c '%s'", idx, cmd, cmd)
```
smlx, line 499: I think this may be improved in a couple of ways:
1. calculating the sha256sum in Go will vastly simplify the shell command. it was only using awk etc. because it was written in Bash before.
2. The command will need to be [shell escaped](https://github.com/alessio/shellescape/blob/853065111c4c5878f9eb9033c262159ecaca0a03/shellescape.go#L23-L41) otherwise it can break if it contains a single quote.
3. not sure if you want to throw a warning or error on duplicate cronjob commands? they'll produce the same sha256sum and so it will be a race to see which starts first.

smlx (commented):
Thanks for picking this up and integrating it with the new Go build-deploy logic!

internal/generator/services.go
shreddedbacon: Happy for you to fix as you think it needs to be

shreddedbacon: We already check for duplicate cronjobs by name, but not command.

tobybellwood (approved):
This is an elegant solution, thanks!

PR #204, opened 2023-05-01
Experimental device requests

rocketeerbkw:


legacy/build-deploy-docker-compose.sh
```diff
@@ -1445,6 +1445,9 @@ do
     fi
   fi
 
+  # handle gpu configuration
```
shreddedbacon, line 1448: We could wrap this experimental feature under a flag like this https://github.com/uselagoon/build-deploy-tool/blob/main/legacy/build-deploy-docker-compose.sh#L845

`ADMIN_LAGOON_FEATURE_FLAG_X` are only settable on the remote cluster, so we have some constraint to where this can be enabled.

```
if [ "$ADMIN_LAGOON_FEATURE_FLAG_EXPERIMENTAL_GPU_SUPPORT" = enabled ]; then
	. /kubectl-build-deploy/scripts/exec-gpu-generation.sh
fi

rocketeerbkw: Do we need a flag at all? Any cluster that doesn't have the k8s plugins installed and GPU nodes available will just fail to schedule pods and cause a deployment failure.

shreddedbacon: Right, I guess that is ok. We can re-evaluate the feature wrapping later. In IO cloud if we ever want to restrict users some how, we will need some sort of flagging system though, problem for another day though.

shreddedbacon:
Do we need this still?

PR #206, opened 2023-06-04
refactor: remove default imagecache registry

shreddedbacon:
This removes the hardcoded imagecache and makes it optionally set using lagoon feature flags.

The `remote-controller` can be configured with the following
```
LAGOON_FEATURE_FLAG_(DEFAULT|FORCE)_IMAGECACHE_REGISTRY=imagecache.amazeeio.cloud
```

Or leveraging it on the project or environment level using global scoped variable
```
LAGOON_FEATURE_FLAG_IMAGECACHE_REGISTRY=imagecache.amazeeio.cloud
```

Additionally, it adds imagecache support to some images that previously didn't have it (busybox images)


legacy/helmcharts/postgres-dbaas/values.yaml
```diff
@@ -4,4 +4,7 @@
 
 environment: ""
 
-readReplicaHosts: ""
\ No newline at end of file
+readReplicaHosts: ""
+
+imageCache: ""
+image: uselagoon/php-8.0-cli
```
tobybellwood, line 10: This image has only got another month left - the 8.2 should just slot in here

```suggestion
image: uselagoon/php-8.2-cli
```

shreddedbacon: Can we figure out a better image to use maybe? We have `amazeeio/alpine-mysql-client` for mysql.

Should we maybe piece together a `uselagoon/database-tools` image instead that combines mysql, postgres, and also mongo tools required for the backup purposes?

shreddedbacon: Also considering the last tag on the `amazeeio/alpine-mysql-client` is [4 years ago](https://hub.docker.com/r/amazeeio/alpine-mysql-client/tags) maybe we should get something we can maintain better?

tobybellwood (approved):
LGTM, nice and clean

PR #211, opened 2023-06-08
Adds git sha scope for variable injection

### LINKED ISSUE #210: GIT_SHA as build variable only/allowing setting of scope
I have some questions regarding the GIT_SHA environment variable.

Currently, we're going this, right https://github.com/uselagoon/build-deploy-tool/blob/38694d1fa18e121a4fbe33546593be2a4fe8f486/legacy/build-deploy.sh#L47

But that's the nuclear option, because -- if I understand it properly -- it's going to be injected as a build arg AND it'll be popped into the env vars for the containers (requiring a restarting all pods).

But what about if I only want to use the git sha as a build variable? I have a client that uses it to configure settings for some application integrations, and this lets the track what data comes from what version of their code they've deployed.
This means something like LAGOON_BUILD_NAME isn't going to work for them, since it's not explicitly tied to a version of the code deployed, just to the deployment (i.e. multiple deployments of the same exact code result in different LAGOON_BUILD_NAMEs)

Could we, perhaps, change the behaviour of this to allow the user to select the scope of the GIT_SHA?
Here https://docs.lagoon.sh/using-lagoon-the-basics/lagoon-yml/#environment_variablesgit_sha we say that if it's `true` it gets injected, but we could change that to also, say, take a scope (so, true/build/runtime/both).
This should be trivial to implement.

I'm keen to hear any ideas that let us achieve something similar - although this change to the GIT_SHA behaviour seems to be only a small tweak, that also brings it in line with how with think about env vars generally.

bomoko:
This PR suggests a way of controlling scope for the GIT_SHA variable, as described in #210 

Closes #210

### DIFF
--- a/legacy/build-deploy.sh
+++ b/legacy/build-deploy.sh
@@ -44,13 +44,27 @@ if [ ! -f .lagoon.yml ]; then
   echo "no .lagoon.yml file found"; exit 1;
 fi
 
+# Here we grab the git sha to populate build and runtime environment variables
+GIT_SHA=`git rev-parse HEAD`
 INJECT_GIT_SHA=$(cat .lagoon.yml | shyaml get-value environment_variables.git_sha false)
-if [ "$INJECT_GIT_SHA" == "true" ]
-then
-  LAGOON_GIT_SHA=`git rev-parse HEAD`
-else
-  LAGOON_GIT_SHA="0000000000000000000000000000000000000000"
-fi
+case "$INJECT_GIT_SHA" in
+  true|both)
+    LAGOON_GIT_SHA=$GIT_SHA
+    LAGOON_BUILD_GIT_SHA=$GIT_SHA
+    ;;
+  build)
+    LAGOON_BUILD_GIT_SHA=$GIT_SHA
+    LAGOON_GIT_SHA="0000000000000000000000000000000000000000"
+    ;;
+  runtime)
+    LAGOON_BUILD_GIT_SHA="0000000000000000000000000000000000000000"
+    LAGOON_GIT_SHA=$GIT_SHA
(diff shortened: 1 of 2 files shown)


legacy/build-deploy.sh
```diff
@@ -44,13 +44,27 @@ if [ ! -f .lagoon.yml ]; then
...
+    ;;
+  runtime)
+    LAGOON_BUILD_GIT_SHA="0000000000000000000000000000000000000000"
+    LAGOON_GIT_SHA=$GIT_SHA
+    ;;
+  *)
+    LAGOON_GIT_SHA="0000000000000000000000000000000000000000"
+    LAGOON_BUILD_GIT_SHA=$LAGOON_GIT_SHA
+    ;;
+esac
```
shreddedbacon, line 67: My only concerns with this are that the `environments.git_sha` field goes from being a boolean, to a string. I think it is ok though.

shreddedbacon:
Maybe it makes sense to just set `LAGOON_BUILD_GIT_SHA` to be the git sha all the time, this way users can consume it if they need to without needing to modify their sha settings when they may only need this value in certain images. as `LAGOON_GIT_SHA` influences an entire build process.

There is also the `LAGOON_BUILD_NAME` that can be consumed as a build argument if required for a unique variable in each build that consumed as required by users if they need something that changes reliably during a build.

shreddedbacon:
Is this still something we need to do?

bomoko:
> Is this still something we need to do?

This is 2 years old now, and I don't see this being the kind of problem that I anticipated. I think let's just close it.
Reopen it if you think it makes sense - but I'm not sure this solves any real problems

PR #224, opened 2023-07-24
fix: regex match for cronjob cleanup

shreddedbacon:
This fixes the regex for cronjob cleanup so that it will remove cronjobs that don't exist.

Previously if I had a cronjob named `drush cron` and then renamed it to be `drush cron different`. The current regex check would mean that both of these cronjobs would remain. `drush cron` would not get cleaned up, when it should.

# current
```
re='\<cronjob-cli-drush-cron\>'
text=cronjob-cli-drush-cron
$ if [[ "$text" =~ $re ]]; then echo match; fi
match
text=cronjob-cli-drush-cron-different
$ if [[ "$text" =~ $re ]]; then echo match; fi
match
```

# fixed
```
re='\<^cronjob-cli-drush-cron$\>'
text=cronjob-cli-drush-cron
$ if [[ "$text" =~ $re ]]; then echo match; fi
match
text=cronjob-cli-drush-cron-different
$ if [[ "$text" =~ $re ]]; then echo match; fi
```


tobybellwood (approved):
LGTM

PR #225, opened 2023-09-13
refactor: prefer the service account token over the older deployer token

shreddedbacon:
Small refactor to the way the service account token is determined. This falls back to the older deployer token that was created prior to kubernetes 1.24, and prefers the service account token that is generated on demand by the build pods.

This reduces the possibility of the following warning appearing in build logs causing confusion
```
Warning: Use tokens from the TokenRequest API or manually created secret-based tokens instead of auto-generated secret-based tokens.
```


bomoko (approved):
MERGE! :)

PR #234, opened 2023-09-28
add `fsGroupChangePolicy: "OnRootMismatch"` to all deployments that use persistent bulk storage

### LINKED ISSUE #235: Add support for OnRootMismatch fsGroupChangePolicy
Workloads using ReadWriteMany volumes with a lot of files (multiple 100k) have long startup times with some storage providers (e.g. [csi-driver-nfs](https://github.com/kubernetes-csi/csi-driver-nfs/)). This happens because without securityContext.fsGroupChangePolicy set, the default for many CSI drivers is to always let the kubelet change the permissions and ownership of all files in a volume ([kubernetes.io](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods)).

Setting `securityContext.fsGroupChangePolicy: OnRootMismatch` will fix the file ownership and permissions only if there is a root mismatch which should only happen during initial provisioning or after a change to the `fsGroup`.

Ideally this is set for all workloads, but only if a feature flag is enabled (opt-in) since this is a breaking change.

Schnitzel:
fixes #235


shreddedbacon:
We have a file that is used by rootless workloads already that this would probably be better placed into, rather than each template individually: https://github.com/uselagoon/build-deploy-tool/blob/main/legacy/rootless.values.yaml

smlx:
This might also need testing on at least the main hyperscalers and provisioners used in CI similar to https://github.com/uselagoon/lagoon/pull/2481#issuecomment-774070068

I wouldn't be surprised if Azure does something exciting and different by default.

shreddedbacon:
> This might also need testing on at least the main hyperscalers, and provisioners used in CI similar to [uselagoon/lagoon#2481 (comment)](https://github.com/uselagoon/lagoon/pull/2481#issuecomment-774070068)
> 
> I wouldn't be surprised if Azure does something exciting and different by default.

Yeah, I asked in the issue how this may impact on other provisioners, as there are so many.

It may be worth wrapping this in a feature flag for injection, and for AIO setting this flag to enabled on clusters using this driver?

tobybellwood:
yes - lets wrap this in a feature flag (FS_ON_ROOT_MISMATCH) for now whilst we look at the alternative NFS systems.

PR #237, opened 2023-09-29
remove unknown spec.pod.metadata.labels from prebackuppods

tobybellwood:
Prior to Kubernetes 1.23 (https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/2885-server-side-unknown-field-validation) and enforced in 1.24, kubectl apply has allowed for kubectl apply to have unknown fields in default operation.

Every build contains:
```
Warning: unknown field "spec.pod.metadata.labels"
```

When upgrading the version of kubectl in this tool past 1.24, builds fail because of the unknown fields.

As these fields aren't present in the CRD, the labels aren't applied to prebackuppods, so they can be safely removed.


tobybellwood:
closing in favour of #125

PR #238, opened 2023-10-03
feat: support cleaning up removed lagoonyml routes

shreddedbacon:
Allow for removed routes in .lagoon.yml and Lagoon API to be removed from the namespace.
This is feature flag wrapped for the moment, and will indicate in a build step if any actions would be performed if it detects any that would be removed. 
Eventually this flag should be deprecated and cleanups should occur normally during a build

To enable it per project/environment, add the following variable as a GLOBAL scope
```
LAGOON_FEATURE_FLAG_CLEANUP_REMOVED_LAGOON_ROUTES=enabled
```
It can also be enabled at the cluster level by setting in the remote-controller
```
LAGOON_FEATURE_FLAG_DEFAULT_CLEANUP_REMOVED_LAGOON_ROUTES=enabled
```

It also has a label selector, so that if a route has been added manually, it can get a label `route.lagoon.sh/remove=false` so that this new build step will ignore it.

This is how it appears in the log when it is in no-op mode
image
And in op mode
image


legacy/build-deploy-docker-compose.sh
```diff
@@ -1626,6 +1626,65 @@ set +x
...
+
+if [ ${#DELETE_INGRESS[@]} -ne 0 ]; then
+  if [ "$(featureFlag CLEANUP_REMOVED_LAGOON_ROUTES)" != enabled ]; then
+    echo ">> Lagoon detected routes that have been removed from the .lagoon.yml or Lagoon API"
+    echo "> You can remove these in the next build by setting the flag 'LAGOON_FEATURE_FLAG_CLEANUP_REMOVED_LAGOON_ROUTES=enabled' as a GLOBAL scoped variable to this environment or project"
+  fi
+  for DI in ${DELETE_INGRESS[@]}
+  do
+    if [ "$(featureFlag CLEANUP_REMOVED_LAGOON_ROUTES)" = enabled ]; then
+      kubectl -n ${NAMESPACE} delete ingress ${SINGLE_ROUTE}
```
shreddedbacon, line 1674: We may need to clean up other things associated to this, like any secrets or certificates

shreddedbacon:
I'd like to get this one updated after #255 is merged, so that this could use the warnings system to call out if routes *could* be removed, and the actions the user could take to clean them up.

PR #242, opened 2023-10-12
fix: route generation step and move the break point

shreddedbacon:
Just a fix to actually apply the templates for routes after they're generated. PR #125 moved a step that would apply templates to after the routes are calculated. This meant that the API was not updated to reflect the routes generated, and the `lagoon-env` config-map was not updated either.

Fix is to just do the apply step before like it would have done before so that the calculation succeeds.

A build log break point is also moved into a more suitable location for image builds


tobybellwood (approved):
looks good!

PR #244, opened 2023-10-17
fix: cronjob cleanup use matching function instead of regex

shreddedbacon:
Just fixes up the cronjob cleanup calculation step to be clearer without using regex pattern matching to ensure cleanups are handled.


tobybellwood (approved):
This is a straightforward change, reording the process, and tightening up the previous regex matching. Tested working in Jenkins setup

Approved.

PR #246, opened 2023-10-25
fix: additional prebackuppod changes

shreddedbacon:
Corrects an issue where the variable name used for the readreplicas was wrong


tobybellwood (approved):
tested working.

Straightforward change - renaming the variable correctly.

tobybellwood (approved):
happy with this - retains previous functionality, with the added bonus of using the correct variable

PR #247, opened 2023-10-26
chore: add test for autogenerated enabled, disabled by environment

shreddedbacon:
just adds a test case for disabling an environments autogenerated routes while global autogenerated routes are enabled


tobybellwood (approved):
test only

PR #248, opened 2023-10-26
fix: cronjob cleanup array

shreddedbacon:
Fixes an issue with the cronjob cleanup array that didn't use an array variable in the loop


tobybellwood (approved):
nice fixup

PR #253, opened 2023-11-03
chore: update compose-go spec to v1.20.0 with lagoon fixes

### LINKED ISSUE #252: Validation false positive for double interpolation in docker-compose
When provisioning in Lagoon, the following variable assignment in the `docker-compose.yml` is reported as invalid, while Docker Compose considers this valid and can process it accordingly.

```
MYVAR1: ${MYVAR1:-${MYVAR2:-defaultvalue}substring}
```

When provisioning in Lagoon, the following error is received:
```
There are issues with your docker compose file that lagoon uses that should be fixed.
```

Please note that this is any modern shell supports double variable interpolation, so supporting this in Lagoon seems a reasonable expectation.

shreddedbacon:
This updates the `compose-go` library to 1.20.0 with the additional fixes required for lagoon.

Unfortunately though, this resulted in additional cases where things people may have done in previous versions of docker compose worked, do not work anymore.

Example, previously people may have defined `<<: *default-volumes` and `<<: *default-user` like so, but under the latest compose-go spec, this is not supported.
```
  php:
    build:
      context: .
      dockerfile: .docker/Dockerfile.php
      args:
        CLI_IMAGE: *project
    <<: *default-volumes
    <<: *default-user
    environment:
      <<: *default-environment
    depends_on:
      - cli
    labels:
      lagoon.type: nginx-php-persistent
      lagoon.name: nginx-php
      lagoon.persistent: /app/docroot/sites/default/files/
```

closes #252


shreddedbacon:
Leaving as draft until we make a call on when to fail builds for invalid docker-compose files or drop our fork

shreddedbacon:
closing in favour of #304

PR #254, opened 2023-11-03
feat: add cpu limit flag

shreddedbacon:
In a similar manner to memory limits #120 , do the same for CPU


smlx:
Please don't do this as it doesn't work the same way as memory limits. CPU requests are enough. See https://home.robusta.dev/blog/stop-using-cpu-limits

shreddedbacon:
Closing this due to other work on resources being done within Lagoon

PR #255, opened 2023-11-03
feat: introduce warning system

shreddedbacon:
This just introduces some basic warning capturing functionality.

It doesn't capture warnings from anything that a user may do during a build, but anything that we decide to capture will easily be able to utilise the functionality.

It would be cool to be able to capture things from a docker image build process, or from a pre/post rollout task, but this would require parsing the logs from these processes, which we currently do not do (but could maybe do in the future?)

This does not require the UI changes to be merged, the UI changes just make the warnings actually show up, otherwise the warning features are ignored.

### DIFF
--- a/legacy/build-deploy-docker-compose.sh
+++ b/legacy/build-deploy-docker-compose.sh
@@ -275,7 +275,7 @@ You can run docker compose config locally to check that your docker-compose file
 fi
 
 if [[ "$DOCKER_COMPOSE_WARNING_COUNT" -gt 0 ]]; then
-  echo "Read the docs for more on errors displayed here :LINK_TO_DOCS:
+  echo "Read the docs for more on errors displayed here https://docs.lagoon.sh/lagoon-build-errors
 "
   echo "##############################################"
   currentStepEnd="$(date +"%Y-%m-%d %H:%M:%S")"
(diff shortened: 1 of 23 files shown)


legacy/build-deploy-docker-compose.sh
```diff
@@ -222,9 +233,61 @@ You can run docker compose config locally to check that your docker-compose file
...
+You can run docker compose config locally to check that your docker-compose file is valid.
+"
+  fi
+  echo "> There are yaml validation errors in your docker-compose file that should be corrected."
+  echo ERR: ${dccOutput}
+  echo ""
+fi
+
+if [[ "$DOCKER_COMPOSE_WARNING_COUNT" -gt 0 ]]; then
+  echo "Read the docs for more on errors displayed here :LINK_TO_DOCS:
```
tobybellwood, line 278: I'll create two short links in https://github.com/uselagoon/lagoon/pull/3608
`https://docs.lagoon.sh/lagoon-build-errors` and
`https://docs.lagoon.sh/docker-compose-errors`
to point to a new page to house these errors

tobybellwood (approved):
I'm happy with this  - works as advertised, and will add great functionality to other aspects of Lagoon BUilds

PR #263, opened 2023-11-15
refactor: add a check for username override key variable from api

shreddedbacon:
This adds a username override key to internal container registry login in a similar manner to how #132 allowed for the password to be overridden.

Given the following container-registries block, and wanting to change the username for the account defined from `robot$robotic` to something else, you can add a project or environment variable of the scope `container_registry` with the name `REGISTRY_my-harbor_USERNAME` and set it to whichever username it should be instead.

```
container-registries:
  my-harbor:
    username: robot$robotic
    password: my-special-password
    url: harbor.example.com
```

With a verify script https://gist.github.com/shreddedbacon/892f8206282aed54c029ed1979b6fcb6


tobybellwood (approved):
Approved, working in cluster

PR #264, opened 2023-11-19
chore: add other missing values for mongo prebackuppods

shreddedbacon:
Just minor work to add the other missing variables for mongo prebackuppods. These aren't actually used by the backup command, but they may get used in the future.

Fixes the backup command to also support mongodb and documentdb backups and dumps the archive which can be restored using `mongorestore`


tobybellwood (approved):
Tested full backup, download, restore process working

PR #265, opened 2023-11-20
refactor: swap dioscuri label for lagoon activestandby label

shreddedbacon:
Once https://github.com/uselagoon/lagoon/pull/3592 has been merged, this PR should also be merged to handle setting, and conversion of dioscuri labels to lagoon activestandby labels

The script contained within the build-deploy-docker-compose.sh could be used outside of a deployment, but only once the build-deploy-tool has received the update. This script could be used to fix ingress in environments if deployments are not able to be performed before other systems start to stop using the older labels/annotations.


tobybellwood (approved):
Approved alongside https://github.com/uselagoon/lagoon/pull/3592

PR #266, opened 2023-11-20
refactor: remove unused stakater annotations and add cleanup

shreddedbacon:
This removes the old `stakater` annotations from ingress objects and replaces some existing labels and annotations with new ones, deprecating old ones..

In fitting with https://github.com/uselagoon/lagoon/issues/3597 this would deprecate the label `lagoon.sh/primaryIngress` in favour of `route.lagoon.sh/primary=true|false`.

It will also replace the existing annotation `monitor.stakater.com/overridePath` with the new annotation `route.lagoon.sh/path-override` containing the same value.

The script contained within the `build-deploy-docker-compose.sh` could be used outside of a deployment, but only once the build-deploy-tool has received the update. This script could be used to fix ingress in environments if deployments are not able to be performed before other systems start to stop using the older labels/annotations.

It would be ideal that existing tooling or controllers that may use these old labels and annotations adopt the new ones, with a fallback to support the old ones for a period of time.


internal/templating/ingress/templates_ingress.go
```diff
@@ -100,26 +100,16 @@ func GenerateIngressTemplate(
...
-			additionalAnnotations["monitor.stakater.com/enabled"] = "true"
-			additionalAnnotations["uptimerobot.monitor.stakater.com/alert-contacts"] = "unconfigured"
-			if lValues.Monitoring.AlertContact != "" {
-				additionalAnnotations["uptimerobot.monitor.stakater.com/alert-contacts"] = lValues.Monitoring.AlertContact
-			}
-			if lValues.Monitoring.StatusPageID != "" {
-				additionalAnnotations["uptimerobot.monitor.stakater.com/status-pages"] = lValues.Monitoring.StatusPageID
-			}
-			additionalAnnotations["uptimerobot.monitor.stakater.com/interval"] = "60"
+			additionalLabels["route.lagoon.sh/primary"] = "true"
```
domcyrus, line 107:
```suggestion
			additionalLabels["lagoon.sh/primaryIngress"] = "true"
```
As we already use this annotation and it is set in many many ingresses I think it would be cool to keep it. If we want to change the annotation we will need to have a way to change the already deployed ingresses and also change the ingress operator

internal/templating/ingress/templates_ingress.go
```diff
@@ -100,26 +100,16 @@ func GenerateIngressTemplate(
...
+		} else {
+			additionalLabels["route.lagoon.sh/primary"] = "false"
 		}
 		if route.MonitoringPath != "" {
-			additionalAnnotations["monitor.stakater.com/overridePath"] = route.MonitoringPath
+			additionalAnnotations["route.lagoon.sh/path-override"] = route.MonitoringPath
```
domcyrus, line 112:
```suggestion
			additionalAnnotations["lagoon.sh/monitoringPath"] = route.MonitoringPath
```
Previously we discussed to use this annotation and this is also used on the ingress operator but if you like to use a different annotation we can certainly change this.

domcyrus (commented):
`lagoon.sh/primaryIngress` is already in use and set on many ingresses. Therefore if we want to change this annoatation we would need to have something like an upgrade path which changes the existing annotation to the new one `route.lagoon.sh/primary` and also update the ingress operator.

shreddedbacon:
> `lagoon.sh/primaryIngress` is already in use and set on many ingresses. Therefore if we want to change this annoatation we would need to have something like an upgrade path which changes the existing annotation to the new one `route.lagoon.sh/primary` and also update the ingress operator.

Thanks for the review. This is still a draft at the moment, and if [we decide to move forward with the label or annotation changes](https://github.com/uselagoon/lagoon/issues/3597), you will need to adjust the ingress operator to handle the new one, or both.

As the description of this PR states, there is also a way to go through and update clusters to change the label on all resources (the script can be easily tweaked to do an entire cluster), in which case the ingress operator would only need to know the new one.

Once we decide on the path forward, we will inform you of the change you need to perform.

domcyrus:
@shreddedbacon I see that you do have an upgrade path but the issue is that the ingress operator is relying already on it. Therefore changing it to the new format without supporting the old one is not feasible as you would need to synchronize the roll out with all the potential pitfalls of doing this changes in sync. 
The second issue is with our dashboards which do also make use of this annotations and which will break as well in a way which is not very easy to fix unfortunately because the data is already in prometheus with the old label and there is no way you can change the data in prometheus/thanos...

PR #267, opened 2023-11-23
fix: resource overrides in nginx-php templates

shreddedbacon:
Just a fix that allows for the resources that get defined by overrides (like container memory limits) to get applied to nginx-php based templates, which didn't happen before because they had container specific definitions that meant the overrides never got applied.


tobybellwood (approved):
Tested working, nice pickup

PR #269, opened 2023-11-24
fix: basic service port override when the service is set

shreddedbacon:
The step where the service is created didn't know about the port change capabilities of the `basic` service type. This fixes it so that it will set the override if it is set.


tobybellwood (approved):
tested working. It might be a weird side-effect of the double apply, but this is good to go.

shreddedbacon:
> tested working. It might be a weird side-effect of the double apply, but this is good to go.

Yeah, it is

PR #271, opened 2023-11-27
refactor: only generate schedules if resource types have backups

### LINKED ISSUE #270: Environments with nothing to backup still create schedules
When an environment is deployed in a completely stateless way (no database, no persistent volumes) the schedule for backups is still created.

This can lead to unexpected issues when the purge and check jobs run for these environments, they will fail if there is no associated restic repository, as this is only created when the first backups are performed.

The build-deploy-tool should check if there are any components that required backups, and if so, only then create the schedule.

shreddedbacon:
Refactoring the schedule generation to only generate if a resource type contains items that can be backed up

# Closing issues

closes #270


internal/generator/services.go
```diff
@@ -35,6 +35,26 @@ var supportedAutogeneratedTypes = []string{
 	"python",
 }
 
+// these are lagoon types that come with resources requiring backups
+var typesWithBackups = []string{
```
shreddedbacon, line 39: Need some eyes on this to make sure that these are the types that have something that requires backups

tobybellwood: looks complete to me - cli-persistent and worker-persistent have no need for backups as they are exclusively using other services storage.

internal/testdata/node/docker-compose.autogen-1.yml
```diff
@@ -16,6 +16,20 @@ services:
       - LAGOON_LOCALDEV_HTTP_PORT=3000
       - LAGOON_ROUTE=http://node.docker.amazee.io
 
+  elasticsearch:
+    image: uselagoon/elasticsearch-7:latest
```
tobybellwood, line 20: I'd have a preference in the tests for using an image we still support (ES is EOL for us) - would the Opensearch image work?

shreddedbacon: well these never get deployed anywhere beyond being in the tests. so im going to change it to something completely unrelated if anything

shreddedbacon: fixed

tobybellwood (approved):
With all tests passing, I'm happy to approve this. Will observe in test cluster

PR #274, opened 2023-12-02
feat: clean up unused services and resources

shreddedbacon:
This adds support for cleaning up unused services within an environment.

As Lagoon has typically not cleaned up in the past, this is a potentially dangerous operation in some cases. A user may be unaware that they may still be using or have an old service in their environment, and their application may still be using it.

There isn't any way to actually confirm if a service is still in used by an application during a build, so the action will be to `warn` and let the user decide on what action to take.

Maybe in the future removals will happen automatically and it will produce a warning still, but no choice in action.

Currently, this can detect and clean up any services, or dangling resources that a service created. In this case a typical service comprises of Kubernetes resources of the types `Deployment`, and then depending on the service, a `Service` and/or a `PersistentVolumeClaim`. A service can also be one of the DBaaS types.

Once the build has detected these, it will output the resources that would be removed and instruct what to do to have them removed. The user can verify the changes and make the decision to remove them in a future build, but also makes it clear that they need to verify that the resources are not used anymore before they are removed.

A new function that will save the service states into a `lagoon-services` configmap which can be used by the [remote-controller](https://github.com/uselagoon/remote-controller/pull/321) to send environment state back to the API for visibility.

To properly support all the changes in this in the API and other systems, the following are required, and the references to pr-99 are added in code to identify where the changes will need to be made in the future to use machinery.
* https://github.com/uselagoon/machinery/pull/99
* changes to remote-controller to source new configmap with services information
* changes to api to store additional data for services

### DIFF
--- a/internal/cleanup/cleanup.go
+++ b/internal/cleanup/cleanup.go
@@ -38,7 +38,7 @@ func RunCleanup(c *collector.Collector, gen generator.GeneratorInput, performDel
 			if performDeletion {
 				fmt.Printf(">> Removing deployment %s\n", i.Name)
 				if err := c.Client.Delete(ctx, &i); err != nil {
-					return nil, nil, nil, nil, nil, nil, err
+					fmt.Printf("!! Error removing deployment %s\n", i.Name)
 				}
 			} else {
 				fmt.Printf(">> Would remove deployment %s\n", i.Name)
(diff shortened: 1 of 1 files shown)


shreddedbacon:
This needs more work before it is ready for a release, so I've removed the milestone

internal/cleanup/cleanup.go
```diff
@@ -0,0 +1,138 @@
...
+		fmt.Println(`> Future releases of Lagoon may remove services automatically, you should ensure that your services are up always up to date if you see this warning."`)
+
+		var mariaDBToDelete, mongoDBToDelete, postgresToDelete, volumesToDelete, servicesToDelete, deploymentsToDelete []string
+		ctx := context.Background()
+		for _, i := range depDelete {
+			deploymentsToDelete = append(deploymentsToDelete, i.Name)
+			if performDeletion {
+				fmt.Printf(">> Removing deployment %s\n", i.Name)
+				if err := c.Client.Delete(ctx, &i); err != nil {
+					return nil, nil, nil, nil, nil, nil, err
```
CGoodwin90: If this deletion fails it `returns`, potentially leaving it in partially cleaned state. Can probably log the `error` and `continue` instead

PR #275, opened 2023-12-11
fix: actually source the namespace

shreddedbacon:
Fix to actually source the namespace from the build for use in templating backup schedules. Backup schedules are the only part of a build step in the tool that actually uses the namespace currently.


tobybellwood (approved):
oh wow - nice fix.

PR #277, opened 2023-12-11
chore: allow images merging against a release branch to build images

shreddedbacon:
This would allow us to create `release/*` branches, then we can build PR images for hotfixes against a particular release tag/version


tobybellwood (approved):
yes - this makes sense

PR #280, opened 2023-12-12
refactor: remove the duplicated namespace calculation for the generator value

shreddedbacon:
Just removing the duplicated namespace calculation step from the tasks runner to use the value that comes from the generator output now that #278 is merged and resolves the namespace being provided to the generator.


bomoko (approved):
Yeah, all makes sense in context. Shipit.

PR #282, opened 2023-12-20
fix: ignore acme challenge ingress in cleanup

shreddedbacon:
Just a minor fix up so that the route cleanup script ignores TLS challenge ingress. This also changes the old `acme.openshift.io/exposer` to the new `acme.cert-manager.io/http01-solver` for existing routes so that challenges are ignored there too.


rocketeerbkw:
Discussed and approved "in person." :+1:

PR #284, opened 2023-12-27
test: consolidate test seed data

shreddedbacon:
Just consolidating the seed test data functions down into reusable packages and data sets for consumption in tests.

It removes a lot of duplicated source `.lagoon.yml` and `docker-compose.yml` files, and moves the results into the testdata directory


cmd/root.go
```diff
@@ -136,6 +136,7 @@ func init() {
 		"Path to the template on disk")
 	rootCmd.PersistentFlags().StringP("saved-templates-path", "T", "/kubectl-build-deploy/lagoon/services-routes",
 		"Path to where the resulting templates are saved")
```
bomoko, line 138: I wonder, as a general thing - whether this needs some kind of documentation?
I understand why the default isn't set here (and is set generator.go, I believe) - but is there any signposts as to what the structure of this argument should be (which would at least be implied by a default value)

bomoko (approved):

internal/generator/ingress.go
```diff
@@ -233,6 +233,7 @@ func autogeneratedDomainFromPattern(pattern, service, projectName, environmentNa
 		domainPart := part
 		if len(part) > 63 {
 			domainPart = fmt.Sprintf("%s-%s", part[:54], domainHash[:8])
+			// domainPart = fmt.Sprintf("%s-%s", part[:54], helpers.GetBase32EncodedLowercase(domainHash)[:8])
```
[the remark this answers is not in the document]
shreddedbacon: :see_no_evil:

cmd/root.go
bomoko: wrong line

cmd/root.go
```diff
@@ -136,6 +136,7 @@ func init() {
...
+	rootCmd.PersistentFlags().String("default-backup-schedule", "", "The default backup schedule to use")
```
bomoko, line 139: I wonder, as a general thing - whether this needs some kind of documentation?
I understand why the default isn't set here (and is set generator.go, I believe) - but is there any signposts as to what the structure of this argument should be (which would at least be implied by a default value)

shreddedbacon: generally, these flags are not used, but they're there as a fallback because the [generator input can use them](https://github.com/uselagoon/build-deploy-tool/pull/284/files#diff-8b340fe7a0d960e9d5c73ae3f5bf7f134235a24779e08f74ea1ae8bb932a8372R105)

Most of the flags are sort of a legacy throw back to the early days of the tool, and might be worth re-evaluating their need in the future.

bomoko: Sweet, makes perfect sense. Just thought I'd raise it :)

shreddedbacon: :100: i'll raise this as an issue and we can look to tackle it in a followup

internal/generator/ingress.go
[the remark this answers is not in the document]
shreddedbacon: gone :smile:

PR #288, opened 2024-01-14
Replace grype with trivy for cyclonedx generation

bomoko:
As it stands, we're using trivy as the sbom scanner in insights-handler, but generating sboms with `syft` on the remote side.

This has lead to issues with format incompatibilities across the two tools - specifically, an [update to the cyclonedx format ](https://github.com/CycloneDX/cyclonedx-go/releases/tag/v0.8.0) wasn't tracked in both tools and lead to a [fatal crash](https://github.com/uselagoon/insights-handler/pull/60)  in the handler.

This brings the insights generation in line, tooling wise, by replacing grype with trivy for sbom generation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -111,9 +111,6 @@ RUN curl -sSL https://github.com/uselagoon/lagoon-linter/releases/download/v0.8.
 #     | tar -xz -C /usr/local/bin build-deploy-tool
 COPY --from=golang /app/build-deploy-tool /usr/local/bin/build-deploy-tool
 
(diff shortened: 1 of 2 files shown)


shreddedbacon (approved):
I'm not super familiar with trivy scanning, but [reading the documentation](https://aquasecurity.github.io/trivy/v0.33/docs/sbom/cyclonedx/) for trivy scanning with cyclonedx, I am satisfied that this PR solves the problem of transferring from syft to trivy.

Issue #287, opened 2024-01-12
Expunge fastly management from build-deploy-tool


shreddedbacon:
This is simply a place holder for documenting the items that need to be removed from the build-deploy-tool. It will probably link out to multiple issues, or in the case of amazee.io(AIO), may be referenced in internal discussions. The fastly control, is as far as we are aware, only used by AIO. If you're a community member of Lagoon that uses any of this in your own hosted Lagoon, please let us know.

* Environment variables - these will need some form of transition away from. Some ofhese are used by AIO, and there will need to be some way for Lagoon and its users to be able to pass this information to the cluster in another way (which has yet to be discussed or decided, as stated this issue is purely to call out the things that need to be removed)
	* `LAGOON_FASTLY_SERVICE_ID`
	* `LAGOON_FASTLY_SERVICE_IDS`
	* `LAGOON_FASTLY_NOCACHE_SERVICE_ID`
	* `LAGOON_FASTLY_AUTOGENERATED`
* `fastly-api-secrets` - The functionality contained in this section of code https://github.com/uselagoon/build-deploy-tool/blob/dfd540684e3be257d6047188533eaa40ab37152e/legacy/build-deploy-docker-compose.sh#L1003-L1107 and https://github.com/uselagoon/build-deploy-tool/blob/dfd540684e3be257d6047188533eaa40ab37152e/legacy/scripts/exec-fastly-api-secrets.sh can/should be removed. This functionality has never been utilised within AIO, and I believe there is no reason to support this functionality from this point on.

PR #289, opened 2024-01-14
feat: replace helm with go service templates

### LINKED ISSUE #208: Branch names consisting of only numbers are not handled correctly
**Describe the bug**
During testing, a branch named `1299` was not picking up custom routes defined in the `.lagoon.yml` correctly. It looks like this issue is actually shyaml's indeterminate typing for a string that can be either an integer or a string. 

**To Reproduce**
1. Create a branch with only numbers as its name, like `1299`
2. Define custom routes for that branch in the `.lagoon.yml` file
4. Observe that the custom routes are not picked up during the lagoon build.

**Expected behavior**
Since a branch name consisting of only numbers _should_ be a valid branch name, Lagoon should support using solely digits as branch names

**Additional context**
The fix for this issue will likely involve forcing shyaml to read every branch name as a string, to remove any ambiguity with regards to the type of the branch name.

### LINKED ISSUE #243: Validate all cronjobs across all environments are valid
Spotted in a deployment:

```
+++ cat .lagoon.yml
+++ shyaml get-value environments.production.cronjobs.3.service
Error: invalid path 'environments.production.cronjobs.3.service', missing key 'service' in struct.
++ CRONJOB_SERVICE=
```

So the cronjob was missing a required field `service`. This fails in step 13 in the deployment, it would be better I think to validate all cronjobs across **all environments** in the linter (which is much earlier in step 4).

In this particular situation, it impacted the production deployment, and was not able to be tested on development environments first. This leads to frustration from a developer experience perspective.

Bonus points would be making the error message friendlier as well.

### LINKED ISSUE #268: Support specifying multiple ports for basic services using labels
Hi!

This is a feature request. We'd like to set multiple ports for a basic container. Currently only one port can be set using label `lagoon.service.port` in docker-compose.yml

We'd like to set an additional port to gain access to the administration interface of one of the images we're running.

Would it be possible to add support for specifying at least one additional port to a service?

shreddedbacon:
This is a fairly substantial pullrequest that moves away from using helm template generation to service templates defined in go.

Most of the old helm chart logic was identical, as such the work in the PR was to try and extract all the individual changes from each of the original templates into their core components. This meant that the template logic could be consolidated down from all the individual helmcharts. The result is that there are now service types defined in code, which can be extended or modified as required, and the template generator should be able to adjust accordingly with minimal changes overall.

The benefits of this are that we now have far better coverage of changes, and which values influence the resulting templates than before with helm. If modifying one value, you can now see how this change flows through to any templates or any other service generation steps.

A big part of this change is to also remove a substantial amount of legacy bash for handling the image build process. All of the old logic for determining build contexts, targets, dockerfiles, pull through images (both from external and private registries) has been moved into go. The result is simpler bash functionality so it focuses on the important parts only (building or pushing images)

# Mutliple ports
This brings additional functionality that allows multiple ports to be opened for services,
Example docker-compose with the `lagoon.service.usecomposeports: true` label. Adding this results in the default port for the service being disabled, and only those ports defined in the `ports` section of that specific service will be created.
```
version: '2'
services:
  custom:
    build:
      context: internal/testdata/basic/docker
      dockerfile: Dockerfile
    user: '10000'
    labels:
      lagoon.type: basic-persistent
      lagoon.persistent: /app/storage
      lagoon.service.usecomposeports: true
    volumes:
      - ./storage:/app/storage:delegated
    ports:
      - "8080"
      - "10009"
```
it is then possible to expose one of these services to a route in `lagoon.yml` file like so. The first port in the list will always be the "default" port for a service defined route, but it is possible to direct a route to a different port by extending the service name.
```
environments:
  main:
    routes:
      - custom:
        - example.com
      - custom-10009:
        - otherport.example.com
```
> Note: while UDP ports can be defined, attempting to create a route to a UDP service will fail.

---
Note: There is additional work to consolidate the template generation that will need to be done after this pullrequest. That would be consolidating all the template generation steps down, but this would be another substantial change to the overall lagoon build structure that needs to be addressed (ie, we template ingress and services at different points of a build, with consolidation efforts, these could all be applied at the same time in one step)

There are still some other helm templates that remain that would be addressed by #115 and #287 

# Closes

closes #208
closes #243 
closes #268


tobybellwood (approved):
I've been through this *A LOT* - and it looks like it's pretty spot-on to the current configuration. Ok for this to merge and head out the door 🥳

internal/servicetypes/solr.go
```diff
@@ -65,11 +66,14 @@ var solr = ServiceType{
 			Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{
 					corev1.ResourceCPU:    resource.MustParse("10m"),
-					corev1.ResourceMemory: resource.MustParse("100M"),
+					corev1.ResourceMemory: resource.MustParse("10Mi"),
```
seanhamlin, line 69: Is this an intended change? Seems like a regression.

shreddedbacon: The previous value was `10Mi`
```
legacy/helmcharts/solr/values.yaml:    memory: 10Mi
```

PR #291, opened 2024-01-15
fix: remove the secret when tls-acme is set to false

### LINKED ISSUE #290: Setting tls-acme=false removes certificate object, but not secret
After https://github.com/uselagoon/build-deploy-tool/issues/227 was addressed, Lagoon will now remove Certificate objects when tls-acme=false. However, the Secret object containing the TLS certificate and key remains. So eventually the certificate expires (and any certificate monitoring generates alerts).

I guess the Secret object should also be deleted when setting tls-acme=false. Maybe after double-checking that the secret object _definitely_ contains an ACME cert.

Deleting the secret immediately also make any problems more discoverable for the user. For example, if they are relying on a valid origin cert they will get errors immediately upon setting tls-acme=false, rather than months later when they have forgotten what they did.

shreddedbacon:
In combination with https://github.com/uselagoon/lagoon/pull/3638, this fixes the behaviour of when `tls-acme: true` is changed to `false` so that it also correctly cleans up the certificate being served.

closes #290

### DIFF
--- a/legacy/build-deploy-docker-compose.sh
+++ b/legacy/build-deploy-docker-compose.sh
@@ -1931,8 +1931,9 @@ TLS_FALSE_INGRESSES=$(kubectl -n ${NAMESPACE} get ingress -o json | jq -r '.item
 for TLS_FALSE_INGRESS in $TLS_FALSE_INGRESSES; do
   TLS_SECRETS=$(kubectl -n ${NAMESPACE} get ingress ${TLS_FALSE_INGRESS} -o json | jq -r '.spec.tls[]?.secretName')
   for TLS_SECRET in $TLS_SECRETS; do
+    echo ">> Cleaning up certificate for ${TLS_SECRET} as tls-acme is set to false"
+    kubectl -n ${NAMESPACE} delete secret ${TLS_SECRET}-tls
     if kubectl -n ${NAMESPACE} get certificates.cert-manager.io ${TLS_SECRET} &> /dev/null; then
-      echo ">> Cleaning up certificate for ${TLS_SECRET} as tls-acme is set to false"
       kubectl -n ${NAMESPACE} delete certificates.cert-manager.io ${TLS_SECRET}
     fi
   done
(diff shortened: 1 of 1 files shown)


legacy/build-deploy-docker-compose.sh
```diff
@@ -1918,8 +1918,9 @@ TLS_FALSE_INGRESSES=$(kubectl -n ${NAMESPACE} get ingress -o json | jq -r '.item
 for TLS_FALSE_INGRESS in $TLS_FALSE_INGRESSES; do
   TLS_SECRETS=$(kubectl -n ${NAMESPACE} get ingress ${TLS_FALSE_INGRESS} -o json | jq -r '.spec.tls[]?.secretName')
   for TLS_SECRET in $TLS_SECRETS; do
+    echo ">> Cleaning up certificate for ${TLS_SECRET} as tls-acme is set to false"
```
domcyrus, line 1921: This doesn't check whether the certificate is actually a `tls-acme`/`LE` certificate. I think this could lead into issues where non `tls-acme`/`LE` certificate gets deleted just because `kubernetes.io/tls-acme` is set to `false`.

shreddedbacon: Do you happen to have a snippet that could verify this?

domcyrus: @shreddedbacon Could we check for a `cert-manager` annotation on the secret first e.g. this ones: 
```text
   cert-manager.io/issuer-group: cert-manager.io
   cert-manager.io/issuer-kind: ClusterIssuer
   cert-manager.io/issuer-name: letsencrypt-prod
```

shreddedbacon: I've already updated the PR to check the issuer directly from the cert using openssl.

Unfortunately, we can't rely on the `letsencrypt-prod` issuer always being called that. Numerous articles online I have seen them tell people to call it `letsencryt-production` or `prod-letsencrypt`. While for AIO cloud we know it is `letsencrypt-prod`, not everyone calls it this.

shreddedbacon:
Just rebased on main, no changes otherwise

PR #292, opened 2024-01-22
Adds ability to control insights scan image with env vars

bomoko:
Currently the insights image is fixed in the `exec-generate-insights-configmap.sh` file - this PR:

- makes use of the globally set IMAGECACHE_REGISTRY env var to set an image cache
- introduces INSIGHTS_SCAN_IMAGE env var to allow overriding of the image used to run the scan

### DIFF
--- a/legacy/scripts/exec-generate-insights-configmap.sh
+++ b/legacy/scripts/exec-generate-insights-configmap.sh
@@ -8,7 +8,7 @@ SBOM_CONFIGMAP="lagoon-insights-sbom-${IMAGE_NAME}"
 IMAGE_INSPECT_CONFIGMAP="lagoon-insights-image-${IMAGE_NAME}"
 IMAGE_INSPECT_OUTPUT_FILE="${TMP_DIR}/${IMAGE_NAME}.image-inspect.json.gz"
 
-# Here we give the user the ability to override the insights scan image
+# Here we give the cluster administrator the ability to override the insights scan image
 INSIGHTS_SCAN_IMAGE="aquasec/trivy"
   if [ "$ADMIN_LAGOON_FEATURE_FLAG_INSIGHTS_SCAN_IMAGE" ]; then
     INSIGHTS_SCAN_IMAGE="${ADMIN_LAGOON_FEATURE_FLAG_INSIGHTS_SCAN_IMAGE}"
--- a/legacy/scripts/exec-generate-insights-configmap.sh
+++ b/legacy/scripts/exec-generate-insights-configmap.sh
@@ -8,6 +8,9 @@ SBOM_CONFIGMAP="lagoon-insights-sbom-${IMAGE_NAME}"
 IMAGE_INSPECT_CONFIGMAP="lagoon-insights-image-${IMAGE_NAME}"
 IMAGE_INSPECT_OUTPUT_FILE="${TMP_DIR}/${IMAGE_NAME}.image-inspect.json.gz"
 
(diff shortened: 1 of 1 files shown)


legacy/scripts/exec-generate-insights-configmap.sh
```diff
@@ -8,6 +8,9 @@ SBOM_CONFIGMAP="lagoon-insights-sbom-${IMAGE_NAME}"
 IMAGE_INSPECT_CONFIGMAP="lagoon-insights-image-${IMAGE_NAME}"
 IMAGE_INSPECT_OUTPUT_FILE="${TMP_DIR}/${IMAGE_NAME}.image-inspect.json.gz"
 
+# Here we give the user the ability to override the insights scan image
+INSIGHTS_SCAN_IMAGE="${INSIGHTS_SCAN_IMAGE:-aquasec/trivy}"
```
shreddedbacon, line 12: There isn't a way for the `INSIGHTS_SCAN_IMAGE` to be initially seeded in anyway, so this will always be `aquasec/trivy`.

Depending on how this could be set, we may need to either allow it to be changed by the remote-controller only (meaning users can't change it) by consuming an `[ADMIN_LAGOON_FEATURE_FLAG](https://github.com/uselagoon/build-deploy-tool/blob/3902ec669048736aa3f2c4434fadd29ca4319f2f/legacy/build-deploy-docker-compose.sh#L920-L922)`

By using the basic `[featureFlag](https://github.com/uselagoon/build-deploy-tool/blob/3902ec669048736aa3f2c4434fadd29ca4319f2f/legacy/build-deploy-docker-compose.sh#L161-L168)` function which allows it to be set at the controller as a default, as a user set variable in a project/env, or as a forced override in the controller.

I'm ok with merging this for now, as I don't think it is a huge problem, but as a minimum we should probably configure `ADMIN_LAGOON_FEATURE_FLAG_INSIGHTS_SCAN_IMAGE` as an override, so that if there is any reason to change it, we don't need to roll an image or bake any support into the API.

bomoko: So I introduced it more as a const, of sorts - really just to indicate that there's a magic-string in there - so minimally it's indicated that there's something there that _can_ or _should_ change.

I was considering even changing the template for the docker call itself (i.e. the bit where it calls `scan`) but thought that was overkill.

Supporting basic feature flag features seems right though.

shreddedbacon: I'm fine with it being the way it is for now, but it was more that if there was a need to be able to change the image or the image version some how, there would be no way to do that currently.

Minimally, the `ADMIN_FEATURE` one would be good though so its at least configurable by remote-controller helmchart only, which is better than having to retag or push a new build image. And clients can't change it either then.

legacy/scripts/exec-generate-insights-configmap.sh
```diff
@@ -8,6 +8,12 @@ SBOM_CONFIGMAP="lagoon-insights-sbom-${IMAGE_NAME}"
 IMAGE_INSPECT_CONFIGMAP="lagoon-insights-image-${IMAGE_NAME}"
 IMAGE_INSPECT_OUTPUT_FILE="${TMP_DIR}/${IMAGE_NAME}.image-inspect.json.gz"
 
+# Here we give the user the ability to override the insights scan image
```
shreddedbacon, line 11: well, not the user, but the cluster administrators

shreddedbacon (approved):

PR #294, opened 2024-01-23
fix: remove any storage calculator pods during a build

### LINKED ISSUE #293: Storage calculator pod causes Lagoon build failure
Consider this series of events:

1. Storage calculator pod starts running to check the storage of a service with a persistent storage volume. e.g. `solr`.
1. A Lagoon build is triggered, causing the `solr` pod to be redeployed.
1. The `solr` pod is scheduled onto a different Kubernetes node than the storage calculator.

The `solr` pod will now be stuck in `ContainerCreating` because it cannot bind the `RWO` volume, which is already attached to the storage calculator pod.

This will cause a Lagoon build failure that the customer cannot influence. In fact, all subsequent builds will fail until one of the following occurs:

* the storage calculator pod completes and exits, freeing the volume binding. This could take several hours.
* the `solr` pod happens to be scheduled onto the same node as the storage calculator.

Here's a sceenshot of this problem occuring. The storage calculator has been running (and blocking any Lagoon builds) for over two hours at this point:
screenshot_2023-12-18-102634

shreddedbacon:
Just extends the build to clean up any storage calculator pods before resources are applied to prevent blocking deployments of workloads.

closes #293

### DIFF
--- a/legacy/build-deploy-docker-compose.sh
+++ b/legacy/build-deploy-docker-compose.sh
@@ -1780,6 +1780,12 @@ set -x
 ### APPLY RESOURCES
 ##############################################
 
+# remove any storage calculator pods before applying deployments to prevent storage binding issues
+STORAGE_CALCULATOR_PODS=$(kubectl -n ${NAMESPACE} get pods -l lagoon.sh/storageCalculator=true --no-headers | cut -d " " -f 1 | xargs)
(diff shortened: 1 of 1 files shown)


tobybellwood (approved):
looks good - no obvious issues, and the logic is sound

PR #295, opened 2024-01-28
refactor: rearrange selectors for flagged ingress

shreddedbacon:
The logic to check the routes for cleanup includes a check for `lagoon.sh/remove`, but this label does not exist unless it has been added to an ingress specifically.

This PR adds a check for this label specifically, and then removes any that have from the discovered routes before the discovered routes are analysed to determine if any ingress need to be removed


tobybellwood (approved):
looks ok to me.

PR #296, opened 2024-01-31
fix: postgres-single backup command use correct variables

shreddedbacon:
Just a quick fix identified through general testing.

The previous backup command for `postgres-single` used a username and password derived from the service name, which is not correct. Only the port can be derived this way due to `enableServiceLinks` which creates the port from the service name. The other variables are defined within the lagoon provided base images.


tobybellwood (approved):
This looks ok to me - can confirm that only the default variable names exist in a single pod

PR #298, opened 2024-02-14
fix: ingress cleanup/warning checks

shreddedbacon:
#295 label selectors were rearranged, this does it again further. The reason is that the `acme` challenge label is not always present either.

#283 the check for the count is never 0, this changes it to check if the value is just not empty instead. This was discovered when checking a namespace that had a valid certificate, the response was still 1.


tobybellwood (approved):
looks ok to me

PR #300, opened 2024-02-25
refactor: use new insights.lagoon.sh/type label

bomoko:
This adds an explicit label used for typing insights ConfigMap data.

It essentially duplicates the data that is currently in the `lagoon.sh/insightsType` label, but brings it in line with [this proposal](https://github.com/uselagoon/lagoon/issues/3597).

Furthermore, by dropping the `-gz` designation in the type, we are pointing towards the proposed introduction of [stricter types for insights-remote](https://github.com/uselagoon/insights-remote/issues/19). With regards to this, note how the new `insights.lagoon.sh/type` does not reference whether the data is compressed or not - this acknowledges that we will only process compressed binary data for `sbom` and `inspect` types going forward.

### DIFF
--- a/legacy/scripts/exec-generate-insights-configmap.sh
+++ b/legacy/scripts/exec-generate-insights-configmap.sh
@@ -44,7 +44,8 @@ processImageInspect() {
       lagoon.sh/buildName=${LAGOON_BUILD_NAME} \
       lagoon.sh/project=${PROJECT} \
       lagoon.sh/environment=${ENVIRONMENT} \
-      lagoon.sh/service=${IMAGE_NAME}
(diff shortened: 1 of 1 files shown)


tobybellwood (approved):
straightforward enough, ties in with rest of insights ecosystem!

shreddedbacon (approved):

shreddedbacon:
Is there an order of release required for this, does it rely on another component to exist or have changes rolled out before this is merged and released, or can it be merged without another system requiring an update?

bomoko:
@shreddedbacon no dependencies :) can be rolled out as is

PR #301, opened 2024-03-05
backport 298 and 286 to pr281

shreddedbacon:
backport 298 and 286 to pr281

for managed cluster, can be closed once 2.18 is released


shreddedbacon:
Not required

PR #302, opened 2024-03-08
refactor: dbaas meta service check

shreddedbacon:
Just refactoring the meta service check for when just `mariadb`, `postgres`, or `mongo` are requested.
The move to go in #115 didn't fully convert the meta checks as some things still take place in BASH for generating the `-single` versions of these types.

### DIFF
--- a/legacy/build-deploy-docker-compose.sh
+++ b/legacy/build-deploy-docker-compose.sh
@@ -389,6 +389,8 @@ if [ ! -z "$LAGOON_ENVIRONMENT_VARIABLES" ]; then
 fi
 set -x
 
+# loop through created DBAAS templates
+DBAAS=($(build-deploy-tool identify dbaas))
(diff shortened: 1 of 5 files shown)


legacy/build-deploy-docker-compose.sh
```diff
@@ -1303,8 +1331,6 @@ do
 
   SERVICE_NAME=${DBAAS_ENTRY_SPLIT[0]}
   SERVICE_TYPE=${DBAAS_ENTRY_SPLIT[1]}
-  # remove the image from images to pull
-  unset IMAGES_PULL[$SERVICE_NAME]
```
shreddedbacon, line 1307: removed from here, because it should be resolved by handling the dbaas checks in the additional code

tobybellwood: correct!

tobybellwood (approved):
Tested working

PR #306, opened 2024-04-17
feat: bump resource requests on stateful pods

smlx:
These resource requests are more realistic based on real-world usage.


tobybellwood (approved):
straightforward change

PR #308, opened 2024-04-24
refactor: make insights gathering step capture as warnings rather than build error

shreddedbacon:
combines #307 and a wrapper around the insights gathering to turn them into warnings


tobybellwood (approved):
Insights errors exit cleanly without impacting the build success. LGTM

PR #310, opened 2024-04-29
refactor: lagoon.yml container-registry variable override checks

shreddedbacon:
Just a quick refactor on the existing variables to allow them to be used without having to define `username` or `password` in the `.lagoon.yml` at all.

[Validation gist](https://gist.github.com/shreddedbacon/25ee93cc4bee89e4510ae0173892eb3d)


tobybellwood (approved):
Tested offline, working

PR #312, opened 2024-05-08
refactor: pod rollout failure log capture

shreddedbacon:
Just refactors the pod rollout failure log capture to capture pods/container logs for the service that are not ready/running.

Previous version used a replicaset hash which didn't always capture the logs for the pods.


tobybellwood (approved):
Straightforward enough change, won't make things worse!

PR #314, opened 2024-05-15
chore: set backup timeout to 14400 seconds

### LINKED ISSUE #313: Backup commands run forever causing excessive disk usage
Redis persistence works [like this](https://redis.io/docs/latest/operate/oss_and_stack/management/persistence/#how-it-works):

> *  Redis forks. We now have a child and a parent process.
> *  The child starts to write the dataset to a temporary RDB file.
> *  When the child is done writing the new RDB file, it replaces the old one.

Meanwhile, the persistent redis backup command in this repository is:

```bash
/bin/sh -c "/bin/busybox tar -cf - -C /data .u
```

---

k8up can sometimes lose track of executed commands such that inside a redis pod you'll see multiple backup commands like this:

```
  426 user      0:00 /bin/busybox tar -cf - -C /data .
  854 user      0:00 /bin/busybox tar -cf - -C /data .
```

When `tar` is executed like this to send data to `stdout` it will open the files in `/data`, bundle them into a file, and then start writing to the `stdout` pipe. Once the `stdout` pipe fills (as can happen if k8up pods restart or otherwise stop reading off the exec pipe), `tar` will be blocked but will retain an open file handle to `/data/restic.rdb`.

Redis will rotate `restic.rdb` files as outlined above, but the disk space will never actually be reclaimed while `tar` holds the open file handle.

This will cause the disk space in `/data` to be filled with old phantom copies of `redis.rdb` and we get disk usage alerts.

I think the best solution will be to set a timeout on the tar command so that it cannot get into the "infinite runtime" state if the stdout pipe fills. Maybe setting the backup command to something like this to enforce a 4 hour maximum runtime?

```bash
/bin/sh -c "timeout 14400 tar -cf - -C /data .
```

tobybellwood:
closes #313

### DIFF
--- a/legacy/helmcharts/redis-persistent/templates/deployment.yaml
+++ b/legacy/helmcharts/redis-persistent/templates/deployment.yaml
@@ -26,7 +26,7 @@ spec:
         {{- end }}
       annotations:
         {{- include "redis-persistent.annotations" . | nindent 8 }}
-        k8up.syn.tools/backupcommand: /bin/sh -c "/bin/busybox tar -cf - -C {{ .Values.persistentStorage.path }} ."
+        k8up.syn.tools/backupcommand: /bin/sh -c "timeout 14400 /bin/busybox tar -cf - -C {{ .Values.persistentStorage.path }} ."
         k8up.syn.tools/file-extension: .{{ include "redis-persistent.fullname" . }}.tar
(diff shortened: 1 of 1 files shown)


shreddedbacon (approved):
Yep, this appears to resolve the issue of processes remaining

PR #315, opened 2024-05-15
build: update golang to 1.22

tobybellwood:
This PR:
* Updates the version of golang used to 1.22
* Updates the upstream images to Alpine3.20
* Updates the kubectl binary to 1.30 and latest helm binary
* Updates most go deps


.github/workflows/build.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v4
       with:
-        go-version: "^1.21"
```
smlx, line 16: Not sure if you are aware but you can use `stable` here to automatically use the latest release. It should be pretty safe due to Go's [compatibility guarantee](https://go.dev/blog/compat).

tobybellwood: we haven't done that, so that we can have a unified strategy - and move all the services to 1.22 at the same time.

shreddedbacon:
What, if any, are the potential impacts of enabling buildkit by default?

tobybellwood:
I _think_ at this stage none, but will obviously have to test thoroughly across a range of scenarios to be sure...

shreddedbacon (approved):
This seems fine to me

PR #316, opened 2024-05-16
chore: lower timeout to 1.5hr for redis-persist backup

shreddedbacon:
#314 set the timeout to 14400 (4hrs) but this may be too long

This lowers it to 1.5hrs. Acknowledging this is pretty hacky solution generally. But if a redis-persistent is taking longer than this to backup, you probably shouldn't be using redis-persistent in a pod.


tobybellwood (approved):
yah - logical

PR #317, opened 2024-05-20
build: initial multiarch commit

tobybellwood:
This PR modifies the dockerfile build process to account for ARM-based systems - either locally or in-cluster

### DIFF
--- a/.github/workflows/docker-image.yaml
+++ b/.github/workflows/docker-image.yaml
@@ -102,8 +102,7 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: .
-          # platforms: linux/amd64,linux/arm64
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
           build-args: |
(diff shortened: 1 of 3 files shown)


.lagoon.yml
shreddedbacon, line 1: lol this file

registry-variable-test.sh
shreddedbacon, line 1: and this one

.github/workflows/docker-image.yaml
```diff
@@ -102,8 +102,8 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: .
-          # platforms: linux/amd64,linux/arm64
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
+          # platforms: linux/amd64
```
shreddedbacon, line 106: do these need to remain commented out, or can they be removed?

shreddedbacon (approved):
All makes sense :shipit: once the test pass.

PR #318, opened 2024-05-22
fix: use jq base64d and remove errant s in cert cleanup

shreddedbacon:
In cases where `tls-acme` is changed from true to false, and a certificate cleanup is required, the step that would check the certificate to see if it was a Let's Encrypt certificate would fail due to the following base64 decode error inside a build pod.

```
>> Cleaning up certificate for example.com-tls as tls-acme is set to false
base64: unrecognized option: decode
```

Initially introduced in #291, this changes the base64 decode to use the one built into JQ like other commands in the bash script do. Also removes an errant `s` in the grep.


tobybellwood (approved):
simple change - makes sense to rely on jq

PR #319, opened 2024-05-28
fix: tls-acme stale challenge check to not use busybox date

shreddedbacon:
Just a small fix to a date check since date in busybox is limited. Replaces it with awk to regex anything greater than 60m/1h instead.

Currently when this process runs, it results in this in the output which can be confusing to users.
```
date: invalid date '1 hour ago'
date: invalid date '2024-05-27T08:53:43Z'
```

This fix also nullifies error output if there are no ingress in the namespace (which is the case majority of the time)


tobybellwood (approved):
makes sense

PR #322, opened 2024-05-31
fix: use yq for checking container-registries because shyaml is not great

shreddedbacon:
shyaml is not great and we should stop using it. As more and more work goes into #289 the usage of shyaml is reduced significantly though.

This changes the checks to use `yq` where possible in the container-registry functions to handle things shyaml isn't great with.

The version of `yq` used is `v4.35.2` which is what is used in the build-deploy-image currently

With updated [gist](https://gist.github.com/shreddedbacon/25ee93cc4bee89e4510ae0173892eb3d) 

#310 introduced the changed logic, but the version of shyaml differs across operating systems and some handled it differently.


tobybellwood (approved):
sensible and correct choice of yq4

PR #324, opened 2024-06-06
chore: increase max-allowed-packet in mysql backup cmds

### LINKED ISSUE #323: max-allowed-packet too small in prebackuppod dump command
I found a backup failure caused by the hard-coded `--max-allowed-packet=500M` in the prebackuppod template being too small. The command was failing with this error:

```
mysqldump: Error 2013: Lost connection to server during query when dumping table `cache_config` at row: 23
```

Bumping the value to `600M` allowed the dump to complete.

The protocol maximum of this value is `1G` in both [RDS](https://aws.amazon.com/blogs/database/best-practices-for-configuring-parameters-for-amazon-rds-for-mysql-part-3-parameters-related-to-security-operational-manageability-and-connectivity-timeout/) and [MySQL](https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_max_allowed_packet). Using that value also worked, so I guess we should just use the maximum value.

screenshot_2024-06-06-085427

The `500M` value is hard-coded in several places. I guess all should be changed?

* https://github.com/uselagoon/build-deploy-tool/blob/997483b59ac7c055a07f04f579b91ccd7e4fb4a2/internal/templating/backups/template_prebackuppod.go#L283
* https://github.com/uselagoon/build-deploy-tool/blob/997483b59ac7c055a07f04f579b91ccd7e4fb4a2/internal/templating/backups/template_prebackuppod.go#L291
* https://github.com/uselagoon/build-deploy-tool/blob/997483b59ac7c055a07f04f579b91ccd7e4fb4a2/legacy/helmcharts/mariadb-single/templates/deployment.yaml#L28

shreddedbacon:
Based on discussion in #323 this just sets the max-allowed-packet to be the maximum supported.

closes #323

### DIFF
--- a/internal/templating/backups/template_prebackuppod.go
+++ b/internal/templating/backups/template_prebackuppod.go
@@ -280,7 +280,7 @@ var preBackupPodSpecs = PreBackupPods{
   BACKUP_DB_HOST=$(echo $BACKUP_DB_READREPLICA_HOSTS | cut -d ',' -f1);
   fi &&
   dump=$(mktemp)
-  && mysqldump --max-allowed-packet=500M --events --routines --quick
+  && mysqldump --max-allowed-packet=1G --events --routines --quick
   --add-locks --no-autocommit --single-transaction --no-create-db
   --no-data --no-tablespaces
   -h $BACKUP_DB_HOST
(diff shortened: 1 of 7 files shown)


smlx:
Will Lagoon overwrite existing deployment annotations and prebackuppod objects in the cluster?

shreddedbacon:
> Will Lagoon overwrite existing deployment annotations and prebackuppod objects in the cluster?

When a deployment occurs, yes

tobybellwood (approved):
Sensible, simple change, LGTM

PR #329, opened 2024-07-01
feat: add feature flag for separate weekly-random prune and check

shreddedbacon:
The current `K8UP_WEEKLY_RANDOM_FEATURE_FLAG` variable impacts both prune and check, so this should be phased out in favour of the new ones that leverage the actual supported featureflag support
```
LAGOON_FEATURE_FLAG(_FORCE|_DEFAULT)_K8UP_WEEKLY_RANDOM_CHECK
LAGOON_FEATURE_FLAG(_FORCE|_DEFAULT)_K8UP_WEEKLY_RANDOM_PRUNE
```

Setting these as `global` scope vars with the value `enabled` turns on the `@weekly-random` schedule for either check or prune

* https://docs.k8up.io/k8up/2.10/references/schedule-specification.html


shreddedbacon:
since this is a relatively harmless change, i'm going to merge it.

most of the minor diffs are just formatting changes, the bulk of the changes just use an OR to check if one flag or the other are set and the functionality of the templating is mostly unchanged

PR #331, opened 2024-07-04
chore: allow docker-host to be configurable

shreddedbacon:
Set the `DOCKER_HOST` in the container image so that it can be modified. Previously it was exported to a value in the entrypoint and unable to be changed.

Also removes the two unused dockerfiles in the legacy to prevent confusion.


legacy/Dockerfile.kubectl-build-deploy-dind
```diff
@@ -19,6 +19,7 @@ COPY scripts /kubectl-build-deploy/scripts
 COPY helmcharts  /kubectl-build-deploy/helmcharts
 
 ENV DBAAS_OPERATOR_HTTP=dbaas.lagoon.svc:5000
+ENV DOCKER_HOST=docker-host.lagoon.svc
```
tobybellwood, line 22: this probably should be set in the root Dockerfile - I think this one isn't built anymore?

shreddedbacon: oh

tobybellwood (approved):
Approved. No tangible changes yet, waiting on other improvements to activate

PR #334, opened 2024-07-19
fix: reduce verbosity in build

shreddedbacon:
This removes the over used `set +x/-x` from the builds reducing the overall verbosity of a build log.


tobybellwood (approved):
Tested working in environments, no significant difference to current output, except less verbosity

PR #335, opened 2024-07-22
fix: correctly consume pr title and cronjob batch v1

shreddedbacon:
This PR fixes up a couple of oversight issues in #289


tobybellwood (approved):
Both issues test ok now.

PR #337, opened 2024-07-22
fix: add LAGOON_CACHE_ build arg support back

shreddedbacon:
This PR restores the LAGOON_CACHE_  functionality into the correct place in the build to allow the ARGS to be used


tobybellwood (approved):
tested working

PR #338, opened 2024-07-23
build: add gha cache for go and docker

tobybellwood:
This PR introduces go and docker caching into the github actions.

In addition, there is now docker layer caching available, so the non-golang layers will also be cached, which can speed up the builds from 12 to 10 minutes.

It uses a hash of go.mod - so if no golang changes are made, the cache is available to the build to use, which can speed up builds from 10 to 2 minutes.


shreddedbacon (approved):
:partying_face: makes this a :pinched_fingers: faster

PR #340, opened 2024-07-26
feat: implement feature to fail builds on docker-compose yaml errors

shreddedbacon:
To help make #304 a reality, this is a phased approach to implement the failure that is detected if a docker-compose file has yaml errors that need to be addressed.

By default, this will make `development` type environments builds fail. It is possible to disable this error using the feature flags
```
# per project/environment
LAGOON_FEATURE_FLAG_DEVELOPMENT_DOCKER_COMPOSE_VALIDATION=disabled

# per cluster in the remote-controller
LAGOON_FEATURE_FLAG_DEFAULT_DEVELOPMENT_DOCKER_COMPOSE_VALIDATION=disabled
```

Production environments will continue to warn, but it is possible to enable the errors using feature flags
```
# per project/environment
LAGOON_FEATURE_FLAG_PRODUCTION_DOCKER_COMPOSE_VALIDATION=enabled

# per cluster in the remote-controller
LAGOON_FEATURE_FLAG_DEFAULT_PRODUCTION_DOCKER_COMPOSE_VALIDATION=enabled
```
Eventually though, the warnings will become failures with no way to disable the error as we upgrade the library.
We will announce before this takes place though.

As warnings have been available in builds for over 6 months now, and these errors are thrown if you're using a new version of docker compose locally, most users will likely have addressed these problems if they had them.

Information about how to temporarily disable the error are presented in the output of the error so that users are able to resolve.

---

This has been flagged for 2.22.0 so that at least for amazeeio there will be time for communication to be provided. The release notes of 2.21.0 should mention this change as coming in 2.22.0 for open source users to be aware of.

### DIFF
--- a/legacy/build-deploy-docker-compose.sh
+++ b/legacy/build-deploy-docker-compose.sh
@@ -255,11 +255,36 @@ dccExit2=$?
 if [ "${dccExit2}" != "0" ]; then
   ((++DOCKER_COMPOSE_WARNING_COUNT))
   if [ "${dccExit}" == "0" ]; then
+
+    # this logic is to phase rollout of https://github.com/uselagoon/build-deploy-tool/pull/304
+    # anything returned by this section will be a yaml error that we need to check if the feature to enable/disable errors
+    # is configured, and that the environment type matches.
+    # eventually this logic will be changed entirely from warnings to errors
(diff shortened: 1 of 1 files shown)


tobybellwood (approved):
This works, and makes sense to implement with itssafeguards

PR #341, opened 2024-07-29
fix: check usespot before applying replicas

shreddedbacon:
Fix the check for adding replicas for production environments when spot is enabled


tobybellwood (approved):
Nice pickup by @rocketeerbkw 

LGTM

PR #344, opened 2024-07-30
fix: polysite cronjob inheritence from lagoon yaml

shreddedbacon:
Polysite cronjob inheritence was broken in the lagoon yaml processor.

This fixes it so that if a polysite project has cronjobs defined, it will still inherit the `environments.$env.cronjobs` and merge them onto the `$project.environments.$env.cronjobs`

Added and updated tests to include this processor for cronjobs on polysites in the lagoon yaml processor


tobybellwood (approved):
test cases mimic real world issue, LGTM

PR #346, opened 2024-07-31
fix: additional templating and logic fixes

### LINKED ISSUE #347: Build arg `LAGOON_SSH_PRIVATE_KEY` not available
It looks like the change to go has dropped the build arg `LAGOON_SSH_PRIVATE_KEY`. It should be set to the project private key, and it's used to facilitate package managers being able to pull from private repos https://docs.lagoon.sh/using-lagoon-advanced/private-repositories/.

shreddedbacon:
fixes an incorrect assumption that all dockerhub images would be from public repositories. Now includes a wrapper that will remove the imagecache injection on dockerhub images if a private repository has been logged into.

pull rate limits will be applied to that user instead of using an imagecache (if an imagecache was supplied)

Also adds a missing build arg for `LAGOON_SSH_PRIVATE_KEY`

Additionally, fixes up a missing or incorrectly associated podsecuritycontext to some services that previously only used `fsGroup: 0`

closes #347

### DIFF
--- a/cmd/identify_imagebuild_test.go
+++ b/cmd/identify_imagebuild_test.go
@@ -602,6 +602,101 @@ func TestImageBuildConfigurationIdentification(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "test10 nginx-php external pull images from dockerhub",
+			args: testdata.GetSeedData(
+				testdata.TestData{
+					Namespace:       "example-project-main",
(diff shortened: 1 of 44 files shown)


tobybellwood (approved):
Tested working as expected

tobybellwood (approved):
updated backup changes make sense too, nice pickup

PR #348, opened 2024-08-01
chore: Add fsGroupChangePolicy to cron jobs

anothertobi:
Deployments already include the fsGroupChangePolicy OnRootMismatch when
enabled. Cron jobs that use the same volumes also need this policy set
to speed up pod startup. This is especially helpful when a volume contains
a large amount of files.


shreddedbacon (approved):
This is just porting the same logic from the deployment template, thank you!

PR #349, opened 2024-08-09
feat: override to be able to force cronjob into service pod

shreddedbacon:
In some cases, a cronjob that would typically be run as a native kuberentes `CronJob` type may be better to run within the pod of the service directly. Since the in pod and native kubernetes switch is based on the interval of the cronjob, there is no way to do this without increasing the frequency of the cronjob, or writing a wrapper around the cronjob.

This offers a simple way to tell a cronjob to run in the service pod instead of a kubernetes native `CronJob` by way of an `inPod` field.

Set this to `true`, and the pod will be converted to an in pod cronjob:
```
- name: my special cronjob
  schedule: "45 * * * *"
  command: example cronjob command
  service: cli
  inPod: true
```


tobybellwood (approved):
THis makes sense to me!

PR #350, opened 2024-08-09
fix: elasticsearch is actually elasticsearch-persistent

shreddedbacon:
Fixes an issue with the template being `elasticsearch-persistent` instead of just `elasticsearch`


tobybellwood (approved):
nice pickup hotfix

PR #352, opened 2024-08-12
fix: better check if spot force is set

shreddedbacon:
Fixes a bug in the logic when trying to check if spot types had the `force` feature enabled on them.


tobybellwood (approved):
Tested working

PR #353, opened 2024-08-13
fix: change duplicate port name in nginx service type

sjerdo:
The nginx service type contains a duplicate port name http which is used by container nginx port 8080 and container php port 9000. This causes a race condition.

The nginx service targets port name nginx, which in some cases may end up good pointing to nginx, but in some other cases this might fail since requests are forwarded directly to the php container.

Before the Go rewrite for Kubernetes templating, port 9000 of the php container was not named


shreddedbacon:
Thanks for this @sjerdo I'm going to close this though in favour of #354 as it has the tests updated. Just want to note your contribution here though.

sjerdo:
Thats fine with me, as long as this bug is fixed

PR #354, opened 2024-08-13
fix: Use distinct container port names

anothertobi:
This introduces the port name "php" for the additional PHP container.

Using the same port name twice can lead to situations where a service
pointing to a named port would route the traffic to the wrong container.


shreddedbacon (approved):
Thanks, this makes sense. Nice catch.

PR #355, opened 2024-08-14
Feature - specify base images to refresh on deploy

bomoko:
Currently, if one chooses to implement a base image solution -- that is, most of one's code being bundled in an upstream image shared by multiple projects/environments -- there is an issue with caching.

If the base image, say, `container.reg/my/baseimage:latest`, or indeed, anything that isn't injected explicitly into the build, the image is cached by the docker-host, and if you make changes to the upstream image, but `container.reg/my/baseimage:latest` is currently _in_ the docker-host's cache, it's not going to pull the new upstream image.

This small change allows users to explicitly mark their services as using a base image, and specifying what the name of that image is via an optional label `lagoon.base.image`.

This leads to, for example, service definitions that look like so:

```
  cli: # cli container, will be used for executing composer and any local commands (drush, drupal, etc.)
    build:
      context: .
      dockerfile: lagoon/cli.dockerfile
    image: &cli-image ${COMPOSE_PROJECT_NAME:-drupal-base}-cli # this image will be reused as `CLI_IMAGE` in subsequent Docker builds
    labels:
      # Lagoon Labels
      lagoon.type: cli-persistent
      lagoon.persistent.name: nginx
      lagoon.persistent: /app/web/sites/default/files/ 
      lagoon.base.image: container.reg/my/baseimage:latest
      lando.type: php-cli-drupal
 <snip/>
```

On finding the label, the build-deploy-tool with attempt to pull the upstream image, which it will do if there have been updates.

### DIFF
--- a/cmd/identify_imagebuild_test.go
+++ b/cmd/identify_imagebuild_test.go
@@ -752,47 +752,6 @@ func TestImageBuildConfigurationIdentification(t *testing.T) {
 				},
 			},
 		},
-		{
-			name: "test11 Force Pull Base Images",
-			args: testdata.GetSeedData(
-				testdata.TestData{
-					Namespace:       "example-project-main",
-					ProjectName:     "example-project",
-					EnvironmentName: "main",
-					Branch:          "main",
-					LagoonYAML:      "internal/testdata/basic/lagoon.forcebaseimagepull.yml",
-				}, true),
-			want: imageBuild{
-				BuildKit: false,
-				BuildArguments: map[string]string{
-					"LAGOON_BUILD_NAME":            "lagoon-build-abcdefg",
-					"LAGOON_PROJECT":               "example-project",
-					"LAGOON_ENVIRONMENT":           "main",
-					"LAGOON_ENVIRONMENT_TYPE":      "production",
-					"LAGOON_BUILD_TYPE":            "branch",
-					"LAGOON_GIT_SOURCE_REPOSITORY": "ssh://git@example.com/lagoon-demo.git",
-					"LAGOON_KUBERNETES":            "remote-cluster1",
-					"LAGOON_GIT_SHA":               "abcdefg123456",
-					"LAGOON_GIT_BRANCH":            "main",
-					"NODE_IMAGE":                   "example-project-main-node",
-					"LAGOON_SSH_PRIVATE_KEY":       "-----BEGIN OPENSSH PRIVATE KEY-----\nthisisafakekey\n-----END OPENSSH PRIVATE KEY-----",
-				},
-				ForcePullImages: []string{
-					"registry.com/namespace/imagename:latest",
-				},
-				Images: []imageBuilds{
-					{
-						Name: "node",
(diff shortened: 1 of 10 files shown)


legacy/build-deploy-docker-compose.sh
```diff
@@ -615,6 +615,14 @@ if [[ "$BUILD_TYPE" == "pullrequest"  ||  "$BUILD_TYPE" == "branch" ]]; then
     BUILD_ARGS+=(--build-arg ${BUILD_ARG_NAME}="${BUILD_ARG_VALUE}")
   done
 
+  # Here we iterate over any lagoon.base.image data that has been passed to us
+  # in order to explicitly
```
shreddedbacon, line 619: Forget to add some words?
```
in order to explicitly pull the images to ensure they are current
```

internal/generator/services.go
```diff
@@ -244,6 +244,11 @@ func composeToServiceValues(
 			}
 		}
 
+		baseimage := lagoon.CheckServiceLagoonLabel(composeServiceValues.Labels, "lagoon.base.image")
+		if baseimage != "" {
+			buildValues.ForcePullImages = append(buildValues.ForcePullImages, baseimage)
```
shreddedbacon, line 249: Just a thought, we're just adding the image here and then the success or failure of this happens during the build step.

Kubernetes does this using validations/reference, we could do that here and drop an error before the build even progresses if the image is bad: https://go.dev/play/p/YNAfkmMLFTh

shreddedbacon: That would catch out a bad image name/tag definition earlier and throw an error then, rather than at the `docker pull` step

bomoko: done :)

legacy/build-deploy-docker-compose.sh
bomoko: would you believe it if I said I was using "explicitly" ironically :P Leaving everything drastically implicit.

shreddedbacon: my brain hurts now lol

cmd/identify_imagebuild_test.go
```diff
@@ -711,6 +711,88 @@ func TestImageBuildConfigurationIdentification(t *testing.T) {
...
+							Context:        "internal/testdata/basic/docker",
+							DockerFile:     "basic.dockerfile",
+							TemporaryImage: "example-project-main-node",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "test11 Force Pull Base Images",
```
shreddedbacon, line 756: Duplicate test

shreddedbacon (changes requested):

cmd/identify_imagebuild_test.go
bomoko: weird. Removed.

shreddedbacon (approved):
This solves the problem neatly, but also in a way that allows for expansion in the future if necessary.

PR #359, opened 2024-08-20
fix: add labels back to cronjobs with a prefix

shreddedbacon:
Due to the way builds apply templates, removing a label is not enough

In https://github.com/uselagoon/build-deploy-tool/commit/fb6924b5e389f9cbabde43f953675be95f0fad61 we removed the label from being set, but in old cronjobs this left the label in place. New cronjobs were not impacted though as they would never have the label.

This changes the behaviour to prefix the labels that would have been set with `cronjob-` so that the values are there, but are not the same as what a deployment would have to prevent them from being counted towards the deployment pods when being selected by SSH services


rocketeerbkw (approved):
Makes sense to resolve issues with environments/cronjobs created before the initial fix

PR #360, opened 2024-08-21
change: update docker and enable buildkit by default

tobybellwood:
This PR:

* Updates the version of docker used in builds to 27.x
* Changes the behaviour to have BuildKit support always-on, but disabled with the  DOCKER_BUILDKIT=0 variable


shreddedbacon (approved):
The changes make sense. If for any reason buildkit enabled by default causes issues, we can update `generator.go` to change the default to `false` easily.

PR #362, opened 2024-08-27
feat: support for path based routing to other services

shreddedbacon:
This adds support for path based routing. This can be used to allow a path on an existing route to be directed to a different service.

For example you have an application that serves a static or other site on `www.example.com` via nginx, but you need to pass `www.example.com/api/v1` to a node application running in the environment. Currently the only way to do that would be via nginx configuration.

This feature allows you to define paths on a route to a different backend.

## Usage

Here is an example `.lagoon.yml` file that shows how path based routes can be defined. There is support for global autogenerated routes, per environment autogenerated routes, and custom routes.
```
docker-compose-yaml: docker-compose.yml

# if an environment has `autogeneratePathRoutes` defined like `environments.main.autogeneratePathRoutes`
# then thes global lagoon yaml `routes.autogenerate.pathRoutes` are ignored
routes:
  autogenerate:
    pathRoutes:
      - from-service: nginx
        to-service: node
        path: /api/v1

environments:
  main:
    # path routes for autogenerated routes for this environment only
    autogeneratePathRoutes:
      - from-service: nginx
        to-service: node
        path: /api/v1
    routes:
      - nginx:
        - a.example.com:
            pathRoutes:
              - to-service: node
                path: /api/v1
```

### Autogenerated Routes

Firstly, there are global autogenerated route configurations, these would apply to ALL environments that this `.lagoon.yml` file would cover.  Second is per environment autogenerated route configurations, the fields are the same, just where they are defined is different.
```
routes:
  autogenerate:
    pathRoutes:
      - from-service: nginx
        to-service: node
        path: /api/v1

environments:
  main:
    # path routes for autogenerated routes for this environment only
    autogeneratePathRoutes:
      - from-service: nginx
        to-service: node
        path: /api/v1
```
> Note: if the per environment path routes are defined on an environment, any global defined ones are ignored.

The 3 fields are
* `from-service` - the autogenerated route for the service you want to add the path based route to
* `to-service` - the backend service you want to route to
* `path` - the path to send to the `to-service`

### Custom Routes

It is possible to turn on path based routes for a specific route too, the following `.lagoon.yml` example shows how.
```
environments:
  main:
    routes:
      - nginx:
        - a.example.com:
            pathRoutes:
              - to-service: node
                path: /api/v1
```
The fields are the same, except that `from-service` is omit because of the way that custom routes are defined to already be associated to a `from-service` in their normal configuration.

### Autogenerated Route configurations

If the global path routes for autogenerated routes are defined, but an environment has overrides, the global defined path routes are ignored by that environment.

### Additional Ports/Docker Compose Ports reference

If the docker compose label `lagoon.service.usecomposeports=true` has been defined on a service, then the service name with the port number as a suffix to any ports beyond the default (first defined port) must be used as the `to-service.

The following `docker-compose.yml` has an example of this label and the ports. This would create the default service port named `node` referencing port 1234, but also another service `node-4321`. If you need to create a path based route to port 4321, you would need to define the `to-service` as `node-4321`.
```
services:
  node:
    build:
      context: .
      dockerfile: basic.dockerfile
    labels:
      lagoon.type: basic
      lagoon.service.usecomposeports: true
    volumes:
      - ./api:/api:delegated
    ports:
      - '1234'
      - '4321'
```

## Closing Issues

closes https://github.com/uselagoon/lagoon/issues/3642


tobybellwood (approved):
This looks good to me, nice fix on the path ports.

PR #363, opened 2024-08-29
feat: new basic-single service type

shreddedbacon:
Introduces a new `basic-single` type.

This is essentially the same as `basic-persistent` except that the volume is ReadWriteOnce (not `bulk` storageclass), and the rollout strategy is changed to `Recreate`.


tobybellwood (approved):
This looks ok to me - we should make sure we document it at some point.

PR #364, opened 2024-09-01
feat: support for pod spread constraints and additional spot controls

shreddedbacon:
This adds some new feature flags to control spot instances, and also a feature flag to set pod antiaffinity rules on pods.

## Pod Topology Spread Constraints
This flag can be used to turn on pod topology spread constraint rules on all deployments. 
*  `LAGOON_FEATURE_FLAG_DEFAULT_POD_SPREADCONSTRAINTS` (this follows the `DEFAULT|FORCE` flags for user controllable feature flags).

The `POD_SPREADCONSTRAINTS` feature will configure pods with constraint rules that will use `whenUnsatisfiable: ScheduleAnyway` to configure all `kind: Deployment` to try and distribute pods across nodes using `topologyKey: "kubernetes.io/hostname"`.

The labels used for the spread constraint rules are the following, and these will be unique to each `kind: Deployment`
* `lagoon.sh/project`
* `lagoon.sh/environment`
* `app.kubernetes.io/name`
* `app.kubernetes.io/instance`

As the rule is also `ScheduleAnyway`, it should still allow for pods to schedule pods on the same nodes as other pods in the event that there is not enough nodes in the cluster to allow for them to be distributed.

## Spot Replicas
Additionally, prior to this feature, all production deployments that were flagged with spot for the types `nginx,nginx-persistent,nginx-php,nginx-php-persistent` were created with 2 replicas. With production remaining the same unless the feature flag is provided as an override.

Now this has been broken out into 2 new feature flags that are only settable on the `remote-controller`.
* `ADMIN_LAGOON_FEATURE_FLAG_SPOT_TYPE_REPLICAS_PRODUCTION`
* `ADMIN_LAGOON_FEATURE_FLAG_SPOT_TYPE_REPLICAS_DEVELOPMENT`

The values of these variables must be the service types that you want to allow to get 2 replicas when running on spot instances.


smlx:
Topology spread constraints might actually be a simpler method to achieve the desired outcome in this case?

https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#comparison-with-podaffinity-podantiaffinity

I kinda forgot about them since topology spread constraints are a newer feature.

shreddedbacon:
> Topology spread constraints might actually be a simpler method to achieve the desired outcome in this case?
> 
> [kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints#comparison-with-podaffinity-podantiaffinity](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#comparison-with-podaffinity-podantiaffinity)
> 
> I kinda forgot about them since topology spread constraints are a newer feature.

Fancy, will read more on it

shreddedbacon:
Unless I'm reading it wrong, the result is basically the same except for the `maxSkew` so I'm not sure how much more simpler it can be than anti-affinity.

Still have to define all the same labels, and the same topologyKey value?

smlx:
I think the main advantages of the newer API are:

* More scalable (anti-affinity has [scalability issues](https://github.com/kubernetes/kubernetes/issues/72479))
* Can add weight via another constraint in future to e.g. prefer nodes from different zones.

shreddedbacon:
Yeah, that sounds fine to me then

tobybellwood (approved):
This looks good to me - may need to monitor some edge cases, but ok to proceed.

PR #365, opened 2024-09-09
fix: support for cron with dual hour ranges

shreddedbacon:
Just some logic fixes for cronjobs with hours that may be multiple ranges.


tobybellwood (approved):
Looks sane to me.

PR #366, opened 2024-09-10
feat: template lagoonenv configmap

shreddedbacon:
Converts the lagoon-env configmap generation into go templating.

Consolidates the wait logic for dbaas consumers into one script as the credential collection step happens outside of this now.

Additionally removes shyaml :partying_face:


shreddedbacon:
Replaced by #397

PR #367, opened 2024-09-10
chore: expunge fastly api secrets templating

shreddedbacon:
This functionality has never been used, and I'd be pretty confident in saying it isn't used by anyone else due to the specifics of how it worked with amazee.io hosting.


tobybellwood (approved):
Good to remove more amazee.io-specific config here

PR #368, opened 2024-09-14
refactor: use robfigcron for validation after lagoon replacements

shreddedbacon:
Just replaces the cron validation logic to use `robfig/cron` to do the final validation. This is what kubernetes uses under the hood in the cronjob controller.

Only the lagoon pseudo random seed replacements are performed and updates the schedule before handing off to be validated.


tobybellwood (approved):
Looks like it simplifies the process a little, and adds standardisation.

The test cases have it well covered.

PR #369, opened 2024-09-24
feat: allow documentation url to be changed

shreddedbacon:
The `remote-controller` has the ability to inject any variables defined against it that start with `LAGOON_FEATURE_FLAG_` into build pods.

This sets the default documentation url to `https://docs.lagoon.sh` but allows for the remote-controller to override this
Adding `LAGOON_FEATURE_FLAG_DEFAULT_DOCUMENTATION_URL` with the url of the alternative documentation, this way references to documentation can point to a version suited for custom installations, as long as the alternative documentation paths are the same.


tobybellwood (approved):
Yup - this is perfect

PR #373, opened 2024-09-30
build: copy required go components to reduce cache invalidation

shreddedbacon:
Just changes up the `COPY . ./` in the dockerfile to copy in Go components only to try reduce cache invalidation


tobybellwood (approved):
Yup - this makes sense

PR #377, opened 2024-10-04
refactor: move pre-rollout tasks to before templating/rollouts

### LINKED ISSUE #375: Run pre-rollout tasks later in the build - ie immediately before the rollout
As discussed with @sjerdo - we should look at moving the pre-rollout tasks closer to the rollout - so that any delays pushing images etc to harbor won't hold up a pre-rollout task - allowing for shorter deployment times (for maintenance mode etc)

shreddedbacon:
This moves the pre-rollout tasks to run before the rollout takes place.

If a pre-rollout task needs to perform some action like putting a site into a mode to prevent traffic while the rollout takes place, running the pre-rollout tasks before the images are built and pushed can mean the site is in an unavailable state longer than needed. Running them after the images are built and pushed results in a shorter period.

The images being pushed does not mean that they are live yet, as this happens during the `Apply Deployments` phase of a build, and pre-rollout tasks will now run between `Image Push` and `Templating Deployments`. If pre-rollout tasks fail, there is no impact to the running environment as no deployment changes will have taken place.

closes #375


legacy/build-deploy-docker-compose.sh
```diff
@@ -377,11 +377,8 @@ declare -A IMAGES_PROMOTE
 # this array stores the hashes of the built images
 declare -A IMAGE_HASHES
 
-HELM_ARGUMENTS=()
+# this sets CAPABILITIES which is used by some processes in this build 
 . /kubectl-build-deploy/scripts/kubectl-get-cluster-capabilities.sh
-for CAPABILITIES in "${CAPABILITIES[@]}"; do
-  HELM_ARGUMENTS+=(-a "${CAPABILITIES}")
-done
```
sjerdo, line 384: Unrelated changes?

legacy/scripts/convert-crontab.sh
sjerdo, line 1: Unrelated changes?

legacy/build-deploy-docker-compose.sh
shreddedbacon: yeah unrelated. these were used when we were still using helm templates

legacy/scripts/convert-crontab.sh
shreddedbacon: yep, just removing a script that has not been used for some time

tobybellwood (approved):
This all makes sense - great improvement on impact!

PR #378, opened 2024-10-06
feat: variable replacement support for lagoon.base.image label

bomoko:
This PR extends #355 by adding the ability to specify tags with environment variables.

It allows the `lagoon.base.image` task to be populated with variables that can be provided to a project or environment and parse them during a lagoon build stage to calculate the required base image to pull.

Variable replacement is similar to `envplate`. They must be in the `${}` format, ie
```
# no default, will error if `BASE_IMAGE_TAG` is not provided by the project or environment
lagoon.base.image: 'repo/image:${BASE_IMAGE_TAG}' 
# default of `main` if `BASE_IMAGE_TAG` is not provided by the project or environment
lagoon.base.image: 'repo/image:${BASE_IMAGE_TAG:-main}'`
# it is possible to replace more than just the tag too
lagoon.base.image: '${BASE_IMAGE_REPO:-repo}/image:${BASE_IMAGE_TAG:-main}'`
```

Nested variables like `${BASE_IMAGE_TAG:-${OTHER_IMAGE_TAG:-latest}}` will not work.

### DIFF
--- a/cmd/identify_imagebuild_test.go
+++ b/cmd/identify_imagebuild_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"testing"
 
+	"github.com/andreyvit/diff"
 	"github.com/uselagoon/build-deploy-tool/internal/dbaasclient"
 	"github.com/uselagoon/build-deploy-tool/internal/generator"
 	"github.com/uselagoon/build-deploy-tool/internal/helpers"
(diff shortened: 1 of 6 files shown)


bomoko:
An alternative might be to forego the `$ENVVAR(:-optionaldefault)` structure entirely and just have multiple labels - like, for instance, a label for a default/fallback separate from one explicitly grabbing the tag from the environment.
We could also try embed the logic in the original refresh image label itself, but parsing becomes complicated because of things like hashes, ports, etc.

shreddedbacon (approved):
After discussing with the team, consensus is that this is fine to merge.

PR #381, opened 2024-10-07
chore: initial DOCKER_PROJECT_BUILDER

tobybellwood:


legacy/build-deploy-docker-compose.sh
```diff
@@ -650,13 +650,43 @@ if [[ "$BUILD_TYPE" == "pullrequest"  ||  "$BUILD_TYPE" == "branch" ]]; then
...
+          echo "Using default builder for $DOCKERFILE";
+        ;;
+        *)
+          echo "Using dedicated builder $PROJECT for $DOCKERFILE";
+          DOCKER_PROJECT_BUILDER=${PROJECT}
+          docker builder create --name $DOCKER_PROJECT_BUILDER --driver docker-container || true
+        ;;
+      esac
+      set -x
+
```
shreddedbacon, line 678: You can condense this entire thing down and use `LAGOON_FEATURE_FLAG_(DEFAULT|FORCE)_DOCKER_PROJECT_BUILDER` rather than some special variable

```
      DOCKER_PROJECT_BUILDER=default
      if [ "$(featureFlag DOCKER_PROJECT_BUILDER)" = enabled ]; then
          DOCKER_PROJECT_BUILDER=${PROJECT}
          docker builder create --name $DOCKER_PROJECT_BUILDER --driver docker-container || true
      fi
```

shreddedbacon: This way someone could enable/disable at the cluster level with either a default `LAGOON_FEATURE_FLAG_DEFAULT_DOCKER_PROJECT_BUILDER` or forcibly with `LAGOON_FEATURE_FLAG_FORCE_DOCKER_PROJECT_BUILDER`

shreddedbacon: Users would set `LAGOON_FEATURE_FLAG_DOCKER_PROJECT_BUILDER=enabled` as a global scope at either project or environment level

tobybellwood: yah - nicer!

legacy/build-deploy-docker-compose.sh
```diff
@@ -668,6 +679,11 @@ if [[ "$BUILD_TYPE" == "pullrequest"  ||  "$BUILD_TYPE" == "branch" ]]; then
   done
 fi
 
+if [ "$(featureFlag DOCKER_PROJECT_BUILDER)" = enabled ]; then  
+  echo "stopping builder $DOCKER_PROJECT_BUILDER"
+  docker builder stop --builder $DOCKER_PROJECT_BUILDER 
+fi
```
shreddedbacon, line 685: what if im running builds across two environments in the same project and my first build stops the builder while the other is still building?

shreddedbacon: This isn't much of a concern considering the failure of being able to share the project builder

PR #384, opened 2024-10-15
fix: check secret exists before trying to read or delete it

shreddedbacon:
Just a quick one to check that the secret of an ingress exists before trying to access it to verify if it needs to be deleted.

Also ensure that the name of the secret is correctly used, not with the `-tls` suffix, as this is already in the variable.


tobybellwood (approved):
Nice easy change, ok to merge

PR #385, opened 2024-10-21
feat: add valkey service type

shreddedbacon:
Adds a supporting `valkey` service type for the recently released Valkey images.

Currently the `redis` and `valkey` types are identical except for the name. But if the projects diverge further, or there are changes specific to Valkey that need to be made, the new type should be used.

This may be difficult for existing users that currently use the `redis` service types since changing types in Lagoon is not currently possible without significant effort. With #274 though, it would be possible to delete a `redis` service and create a new one called `valkey`. Persistent versions using this would experience data loss.

### DIFF
--- a/internal/servicetypes/valkey.go
+++ b/internal/servicetypes/valkey.go
@@ -1,0 +1,101 @@
+package servicetypes
+
+import (
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+var defaultValkeyPort int32 = 6379
+
+var valkey = ServiceType{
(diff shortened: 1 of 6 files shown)


internal/servicetypes/valkey.go
```diff
@@ -0,0 +1,95 @@
...
+				Requests: corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse("10m"),
+					corev1.ResourceMemory: resource.MustParse("10Mi"),
+				},
+			},
+		},
+	},
+}
+
+var valkeyPersistent = ServiceType{
```
shreddedbacon, line 75: Noting that the images state that `VALKEY_FLAVOR=persistent` is required for a persist version of the valkey image to be persistent. This would mean this template should probably to set this envvar?

tobybellwood: yes - that was my plan!

tobybellwood (approved):
Works as expected, thanks!

PR #390, opened 2024-11-18
refactor: split build and push to independent steps

shreddedbacon:
Splits image build and pushes into individual steps where possible to aid in providing more up to date build logs, and help identify where slowness could be occurring during builds.

### DIFF
--- a/legacy/build-deploy-docker-compose.sh
+++ b/legacy/build-deploy-docker-compose.sh
@@ -616,8 +616,6 @@ done
 
 currentStepEnd="$(date +"%Y-%m-%d %H:%M:%S")"
 patchBuildStep "${buildStartTime}" "${buildStartTime}" "${currentStepEnd}" "${NAMESPACE}" "registryLogin" "Container Registry Login" "false"
-previousStepEnd=${currentStepEnd}
-beginBuildStep "Image Builds" "buildingImages"
 
 
 ##############################################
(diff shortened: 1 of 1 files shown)


legacy/build-deploy-docker-compose.sh
```diff
@@ -653,19 +651,23 @@ if [[ "$BUILD_TYPE" == "pullrequest"  ||  "$BUILD_TYPE" == "branch" ]]; then
...
     # if there is no dockerfile, then this image needs to be pulled from somewhere else
     if [ $DOCKERFILE == "false" ]; then
       PULL_IMAGE=$(echo "$IMAGE_BUILD_DATA" | jq -r '.imageBuild.pullImage // false')
       if [ "$PULL_IMAGE" != "false" ]; then
         IMAGES_PULL["${SERVICE_NAME}"]="${PULL_IMAGE}"
+        HAS_PULLED_IMAGES=true
       fi
     else
+      previousStepEnd=${currentStepEnd}
+      beginBuildStep "Building Image ${SERVICE_NAME}" "buildingImage${SERVICE_NAME_TITLE}"
```
tobybellwood, line 670: Is having each buildstep for each image (and push images) individually named going to make it harder to track and compare historic statistics ie
`lagoon.sh/buildStep=buildingImageweb` and `lagoon.sh/buildStep=pushingImagemongo-4`

Is there a way to maintain the current buildingImages step as a wrapper for all images, and still track the individuals?

shreddedbacon: Maybe. However, build steps are not something that should be relied on for statistics as they are always subject to change being a free text field and not an actual build status like `complete` etc..

If we want to enforce these as a method for statistics, we should define what this statistical analysis methods should look like, and it should be saved in the API as part of deployments, rather than as arbitrary labels on deployments.

tobybellwood: yah - fair point - happy to retire this suggestion, but maybe a suggestion here just to split the step and service names with a _ or - for clarity (and same in pushingImage)
```suggestion
      beginBuildStep "Building Image ${SERVICE_NAME}" "buildingImage_${SERVICE_NAME_TITLE}"
```

shreddedbacon: These unfortunately have to somewhat conform with [status conditions](https://github.com/uselagoon/remote-controller/blob/e7c023d27489cb4771aaf4c8c4608f79abaf0f9c/internal/controllers/v1beta2/podmonitor_buildhandlers.go#L560) in kubernetes/remote-controller

We'd need to update the remote-controller to transform these if we include `-` or `_` in the build-deploy-tool. I'm happy to do this, but it would delay this PR.

tobybellwood: ah - ok for now then - but given a service name with a `-` in it, it does come out like so `lagoon.sh/buildStep=pushingImagemongo-4` - is that likely to break anything?

shreddedbacon: :hidethepain: maybe

shreddedbacon: Keep this one on hold, maybe we put it for 2.24 instead with a slight revisit.

shreddedbacon: With the change in `remote-controller` now [modifying the build step](https://github.com/uselagoon/remote-controller/pull/276/files#diff-2978e5518554df13b75f851fd8f3616ed9773bfbc5f361b892cacdee8180f56fR316-R334) as require to conform to k8s status conditions, we should be ok to proceed with this.

Although, maybe updating the buildstep to `pushingImage-mongo-4`, including a `-` before the image name being built. Just to make it obvious where the separation is due to the casing of the step.

PR #393, opened 2024-11-22
fix: add resources limits to linked container as required

### LINKED ISSUE #391: Secondary linked container doesn't respect resource limits
It looks like the code in https://github.com/uselagoon/build-deploy-tool/blob/main/internal/templating/services/templates_deployment.go only applies limits to the first container in a pod, and not the second. This results in e.g. an `nginx-php` pod having no limits applied to the `php` container.

screenshot_2024-11-22-164846

shreddedbacon:
Fixes the missing limits on linked containers

closes #391


tobybellwood (approved):
Makes sense to me - test coverage here is enough to assert correct functionality

PR #395, opened 2024-11-24
chore: remove open/elasticsearch init container

shreddedbacon:
With https://github.com/uselagoon/lagoon-charts/pull/700 we should be able to remove the `initContainer` that sets sysctl values with privileged access. 

The daemonset in the chart will do it instead, best yet would be the underlying k8s nodes having this value increased by default though.

### DIFF
--- a/internal/servicetypes/elasticsearch.go
+++ b/internal/servicetypes/elasticsearch.go
@@ -3,7 +3,6 @@ package servicetypes
 import (
 	"fmt"
 
-	"github.com/uselagoon/build-deploy-tool/internal/helpers"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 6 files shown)


tobybellwood (approved):
Makes sense to me, safe to remove

shreddedbacon:
To remove the initcontainer from the deployments, the following can be used for opensearch and elasticsearch (just change the service-type selector)
```
for dep in $(kubectl get deployments --no-headers -A -l lagoon.sh/service-type=opensearch-persistent -o json | jq -c '.items[] | @base64')
do
    depdata=$(echo "${dep}" | jq -rc '@base64d')
    namespace=$(echo ${depdata} | jq -r '.metadata.namespace')
    deployment=$(echo ${depdata} | jq -r '.metadata.name')
    kubectl -n ${namespace} patch deployment ${deployment} --type json -p='[{"op": "remove", "path": "/spec/template/spec/initContainers/0"}]'
done
```

PR #396, opened 2024-11-25
chore: remove slow capabilities check

shreddedbacon:
This removes the check against the kubernetes API to determine all available APIs. This check can be slow sometimes which can lead to builds being slow in the phase where it collects this information.

Rather than query for everything, target the specific APIs and check for an error if they exist or not to determine the capability exists or not.


tobybellwood (approved):
This looks ok to me

PR #397, opened 2024-11-25
feat: template lagoon-env secret

### LINKED ISSUE #136: Deleted variables from API remain in configmap
**Describe the bug**

When deleting an environment variable from the API, it doesn't remove it from the `lagoon-env` configmap. This is a problem when using the existence of the variable to enable/disable a feature. Like `BASIC_AUTH_USERNAME` and `BASIC_AUTH_PASSWORD`.

**To Reproduce**

1. Add `global|runtime` variable to the API
2. Deploy environment
3. See variable is added to the config map
4. Delete variable from the API
5. Deploy environment
6. See that deleted variable is still in the config map

**Expected behavior**

Deleted API variables should be removed from the configmap

**Impacted customers**

At least 2 large customers in APJ are impacted by this.

### LINKED ISSUE #392: lagoon-env configmap can contain sensitive data
The `lagoon-env` `configmap` deployed by Lagoon can contain sensitive data.

Instead of a `configmap` object this could be a `secret`, providing additional protection for `lagoon-env` data.

shreddedbacon:
# Description

## Secrets :lock: 

The `lagoon-env` configmap generation has been converted into Go templates while also changing to secrets to address #392

## Actually deleting variables?!

This also fixes the long running #136 bug by converting the old `lagoon-env` configmap into two secrets and working out what should and shouldn't exist in each.

* `lagoon-env` - will contain only variables that are in the Lagoon API. 
  * this is the classic `lagoon-env` configmap you're all familiar with, except now it is a secret, and now it only contains what is in the Lagoon API with `global` and `runtime` scoped variables.
* `lagoon-platform-env` - will contain anything that was previously in the configmap, but isn't in the Lagoon API.
  * this secret also serves a second purpose, and is where platform owners could add any platform specific override variables as a way to differentiate them from user added

This allows an environment to retain its current variables without removing any that could result in a change in the operation of the environment. If a user does end up adding a variable that is in the `lagoon-platform-env` secret, it will be removed from that secret, and added to the main `lagoon-env` secret. This way, if a user ends up removing that variable from the API, the cleanup process will now delete it so it won't exist in either secret.

The order that the secrets are added to the deployments is that `lagoon-platform-env` is first, then `lagoon-env`. 
```
envFrom:
- secretRef:
    name: lagoon-platform-env
- secretRef:
    name: lagoon-env
```
This means that variables in `lagoon-env` should always win over any variables in the `lagoon-platform-env` secret (based on ordering discussions in https://github.com/uselagoon/lagoon/pull/2348), this is why the behaviour of removing from the `lagoon-platform-env` secret if a matching variable is found in the Lagoon API is done. Lagoon API wins.

## Minor changes

* Consolidate the wait logic for dbaas consumers into one script, the credential collection step happens outside of this now. This is done to be able to pass these values to the build-deploy-tool for processing in Go
* Finally remove shyaml :partying_face: 

# Closing issues

closes #392 
closes #136 

# Requires

The following should be merged, and ideally released PRIOR to this being merged. Failure to have the below released, before this is merged will result in missing data in build updates, and failure to read the database sizes

* https://github.com/uselagoon/storage-calculator/pull/20
* https://github.com/uselagoon/remote-controller/pull/280


shreddedbacon:
The following controllers leverage the `lagoon-env` configmap too
* remote-controller
* storage-calculator
They will need to have updates to deal with the transition of the configmap to secret(s)

bomoko (approved):
Okay - I've worked through all the code, and I've tested.
It's a _big_ change, but I'm also pretty confident that even if something does slip through the cracks, rolling this back in certain cases isn't actually a problem - the bit that this targets should actually only really be responsive to vars that are in the API (although we know that platform teams do add vars directly to the configmaps).

shreddedbacon:
> rolling this back in certain cases isn't actually a problem

Yeah, the rollback would be that the `lagoon-env` and `lagoon-platform-env` secrets would remain, with the `lagoon-env` configmap being recreated. The only missing part would be that anything in the `lagoon-platform-env` secret would need to be added to the configmap to revert to the previous state.

bomoko:
> > rolling this back in certain cases isn't actually a problem
> 
> Yeah, the rollback would be that the `lagoon-env` and `lagoon-platform-env` secrets would remain, with the `lagoon-env` configmap being recreated. The only missing part would be that anything in the `lagoon-platform-env` secret would need to be added to the configmap to revert to the previous state.

Yes, I think that's doable in an ad hoc way. I think the most important thing to do is to audit clusters for any arbitrary use of the lagoon environment configmap before applying.

shreddedbacon:
This is a snippet that could be used to add values that are in the lagoon-platform-env back into the lagoon-env configmap if reverting would be required.

```
for ENV in $(kubectl -n "$NAMESPACE" get secret lagoon-platform-env --output=json | jq -c '.data|map_values(@base64d)|to_entries|.[] | @base64')
do
    KEY=$(echo "${ENV}" | jq -rc '@base64d' | jq -rc '.key')
    VALUE=$(echo "${ENV}" | jq -rc '@base64d' | jq -rc '.value')
    kubectl -n "$NAMESPACE" patch configmap lagoon-env --type=json -p='[{"op": "add", "path": "/data/'$KEY'","value": "'$VALUE'"}]'
done
```

PR #406, opened 2024-12-20
feat: add support for organization environment variables

rocketeerbkw:
Adds support for organization environment variables in the legacy and golang parts. Also some cleanup of feature flags that were already moved to golang.

Part of https://github.com/uselagoon/lagoon/issues/3602


rocketeerbkw:
There is overlap with #397. I think for the most part, the bits that change the merging logic and most of the tests don't conflict. The short part in the legacy scripts that adds org vars to the config map should be easy to translate.

rocketeerbkw:
We're implementing it in another way.

PR #407, opened 2025-01-16
Chore: Updates Go to 1.23

CGoodwin90:
Updates Go to 1.23


go.mod
```diff
@@ -1,30 +1,30 @@
...
+toolchain go1.23.3
 
 require (
 	dario.cat/mergo v1.0.1
-	github.com/PaesslerAG/gval v1.2.2
+	github.com/PaesslerAG/gval v1.2.4
 	github.com/amazeeio/dbaas-operator v0.3.0
 	github.com/andreyvit/diff v0.0.0-20170406064948-c7f18ee00883
-	github.com/compose-spec/compose-go v1.2.7
+	github.com/compose-spec/compose-go v1.20.2
```
shreddedbacon, line 12: Can't do this, the latest version we can theoretically go to is v1.12.2 (https://github.com/compose-spec/compose-go/blob/v1.12.0/go.mod#L19) as it still uses `gopkg.in/yaml.v2` . The reason we can't go beyond this is because of the way it errors out on specific yaml structures (https://github.com/uselagoon/build-deploy-tool/pull/304).

shreddedbacon: Ahh, yeah we can't actually upgrade this package because of this https://github.com/uselagoon/build-deploy-tool/pull/407/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R93

You'll need to revert this back to 1.2.7 or this `github.com/shreddedbacon/compose-go v0.0.0-20220616064547-4e908a2865c1` needs to be updated in step.

shreddedbacon (approved):

PR #408, opened 2025-01-24
fix: remove unused remnants of custom rollout support

rocketeerbkw:
We used to support custom deployment templates and rollouts under Openshift. I believe this code is some leftovers of that era that was not removed as part of an [old cleanup commit](https://github.com/uselagoon/lagoon/commit/3fdb17baee346f967bafd789f4992daca9cf53b1).

While it is still technically possible for a project to set the rollout type to `false`, and disable rollout monitoring for that service, I can't think of a valid reason why we should continue to support this.

Already have a PR to document it as deprecated https://github.com/uselagoon/lagoon/pull/3871.


shreddedbacon (approved):
Makes sense

PR #409, opened 2025-01-29
feat: environment collection

shreddedbacon:
This is a starting point for being able to inspect a namespace for resources that are or were created by Lagoon at some point.

Sort of a discovery point that can be used in later steps to determine changes that are required, etc.

It is initially set up to do nothing so there would be no impact to merging this as the code in this pullrequest is not executed, as it is all new functionality.


rocketeerbkw (approved):
Approving to merge as a dependency for future features.

PR #410, opened 2025-02-06
fix: use LAGOON_ENVIRONMENT_VARIABLES as primary source

rocketeerbkw:
In https://github.com/uselagoon/lagoon/pull/3856, Lagoon core will consolidate env vars before sending to Lagoon remote, so LAGOON_PROJECT_VARIABLES may be empty.

### DIFF
--- a/legacy/build-deploy.sh
+++ b/legacy/build-deploy.sh
@@ -11,10 +11,10 @@ REGISTRY=$REGISTRY
 NAMESPACE=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
 LAGOON_VERSION=$(cat /lagoon/version)
 
-if [ ! -z "$LAGOON_PROJECT_VARIABLES" ]; then
-  INTERNAL_REGISTRY_URL=$(jq --argjson data "$LAGOON_PROJECT_VARIABLES" -n -r '$data | .[] | select(.scope == "internal_container_registry") | select(.name == "INTERNAL_REGISTRY_URL") | .value' | sed -e 's#^http://##' | sed -e 's#^https://##')
-  INTERNAL_REGISTRY_USERNAME=$(jq --argjson data "$LAGOON_PROJECT_VARIABLES" -n -r '$data | .[] | select(.scope == "internal_container_registry") | select(.name == "INTERNAL_REGISTRY_USERNAME") | .value')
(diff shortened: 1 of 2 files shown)


legacy/build-deploy.sh
```diff
@@ -11,10 +11,10 @@ REGISTRY=$REGISTRY
 NAMESPACE=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
 LAGOON_VERSION=$(cat /lagoon/version)
 
-if [ ! -z "$LAGOON_PROJECT_VARIABLES" ]; then
-  INTERNAL_REGISTRY_URL=$(jq --argjson data "$LAGOON_PROJECT_VARIABLES" -n -r '$data | .[] | select(.scope == "internal_container_registry") | select(.name == "INTERNAL_REGISTRY_URL") | .value' | sed -e 's#^http://##' | sed -e 's#^https://##')
-  INTERNAL_REGISTRY_USERNAME=$(jq --argjson data "$LAGOON_PROJECT_VARIABLES" -n -r '$data | .[] | select(.scope == "internal_container_registry") | select(.name == "INTERNAL_REGISTRY_USERNAME") | .value')
-  INTERNAL_REGISTRY_PASSWORD=$(jq --argjson data "$LAGOON_PROJECT_VARIABLES" -n -r '$data | .[] | select(.scope == "internal_container_registry") | select(.name == "INTERNAL_REGISTRY_PASSWORD") | .value')
+if [ ! -z "$LAGOON_ENVIRONMENT_VARIABLES" ]; then
```
shreddedbacon, line 14: These are set in the `remote-controller` and are added to the project variables.

The API used to set these, but since we stopped using harbor in core, they are no longer provided by the API.

Since we can't rely on the controller being updated before the build image being updated, we'd need to adjust this to check both project and environment variables in a similar way to how you used the `apiEnvVarCheck` function, and then we can work on changing the [remote-controller code to use the environment variables](https://github.com/uselagoon/remote-controller/blob/347cd933e06ac5f91ed5e12776dc792b7f8a0590/internal/controllers/v1beta2/build_helpers.go#L620C25-L628) instead of project.

rocketeerbkw: So we can also fully deprecate that enum in the API?

shreddedbacon: we should probably nuke any columns that contain it in the api-db too

shreddedbacon (approved):
This looks fine to me. Incorporating the backwards compatible project variable check on the necessary variables solves the main concern.

As this is compatible with the current version of lagoon with https://github.com/uselagoon/lagoon/pull/3856 not yet merged (at the time of review) this is fine to be merged without the other being merged.

PR #412, opened 2025-02-10
fix: use docker library to create registry secret authconfig

shreddedbacon:
Fixes a bug in the registry secret generation that doesn't properly escape the payload. Now uses the docker authconfig type to create and json marshal function to convert to byte array for the secret.


bomoko (approved):
Makes sense.

PR #413, opened 2025-02-19
refactor: add retry workaround for image builds

shreddedbacon:
This adds a simple retry logic to capture a weird buildkit related layer issue. While we're working to resolve the actual issue and understand the root cause, this retry logic should help to reduce the instance of failures encountered by forcing a subsequent build to build with the `--no-cache` flag.


tobybellwood (approved):
Tested working

PR #414, opened 2025-02-19
fix: use the k8upv2 feature flag correctly

shreddedbacon:
This fixes a bug with the check performed on which version of k8up should be used when adding annotations. Now correctly uses the feature flag.


tobybellwood (approved):
Looks ok to me!

PR #415, opened 2025-03-01
fix: use backupcommand for basic-single volumes

shreddedbacon:
Fixes the basic-single backups to use `backupcommand` to backup the volume because of RWO volume type.


tobybellwood (approved):
Noice!

PR #416, opened 2025-03-06
fix: add empty platform-env secret as required

shreddedbacon:
Fixes a bug on new environment deployments when checking for the `lagoon-platform-env` secret. Now that https://github.com/uselagoon/remote-controller/pull/280 is merged and released, the remote-controller no longer seeds the initial`lagoon-env` configmap (or the secrets), and builds would fail because the `lagoon-platform-env` secret would never be created.

This introduces an initial check and creation of this secret.


tobybellwood (approved):
makes sense in a k8s-y way

PR #418, opened 2025-03-06
fix: remove unused values file and consolidate service config phase

shreddedbacon:
With the introduction of #397 we removed the need for the values.yaml file that was being populated. It should have been removed in that pullrequest but was not. This removes that file now, and also consolidates what were previously two separate phases into one now that the logic is much simpler.


tobybellwood (approved):
Tested working

PR #424, opened 2025-03-19
fix: use secretKeyRef in prebackuppods

shreddedbacon:
In #397 we changed the `lagoon-env` configmap to secret. This fixes a missing change to prebackuppods that also sets this.


tobybellwood (approved):
looks like the fix to me, happy!

PR #426, opened 2025-03-24
feat: support username password in http giturls

shreddedbacon:
This adds support for being able to define git repository credentials in environment variables for repositories that use HTTP(S) to pull, instead of SSH.

Usage as follows for github, only requiring the `USERNAME` and `PASSWORD` to be defined. These variables must be `build` scope.
```
GITREPO_GITHUB_USERNAME=ghuser1
GITREPO_GITHUB_PASSWORD=ghsomep@ssword
```

To use any other repository, you can create the following, where the repository key, in this example being `EXAMPLE` is used to set the `URL` to use for that repository, with the `USERNAME` and `PASSWORD` variables requiring that same key.  These variables must be `build` scope.
```
GITREPO_EXAMPLE_URL=https://example.com
GITREPO_EXAMPLE_USERNAME=user1
GITREPO_EXAMPLE_PASSWORD=somep@ssword
```

The aim is to support multiple credentials in the event that a submodule that might be in a different repository is also private, and using HTTP(S).

### DIFF
--- a/cmd/template_gitcredential.go
+++ b/cmd/template_gitcredential.go
@@ -1,0 +1,80 @@
+package cmd
+
+import (
+	"fmt"
+	"net/url"
+	"os"
+
+	"github.com/spf13/cobra"
+	generator "github.com/uselagoon/build-deploy-tool/internal/generator"
+	"github.com/uselagoon/build-deploy-tool/internal/helpers"
+	"github.com/uselagoon/build-deploy-tool/internal/lagoon"
+)
+
+var templateGitCredential = &cobra.Command{
+	Use:     "git-credential",
+	Aliases: []string{"gitcred", "gc"},
+	Short:   "Create a git credential file",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		credFile, err := cmd.Flags().GetString("credential-file")
+		if err != nil {
+			return fmt.Errorf("error reading credential-file flag: %v", err)
+		}
+		gitURL, err := TemplateGitCredential(credFile)
+		if err != nil {
+			return err
+		}
+		fmt.Println(gitURL)
+		return nil
+	},
+}
+
+// TemplateGitCredential creates a git credential secret if required
+func TemplateGitCredential(file string) (string, error) {
+	// get the source repository from the build pod environment variables
+	sourceRepository := helpers.GetEnv("SOURCE_REPOSITORY", "", false)
+
+	// because this runs before anything has been checked out in the repo, we have to query variables directly
+	// and handle any merging that is usually done in generator
+	variables := generator.GetLagoonEnvVars()
+
+	// parse the repository into a url struct so the username and password can be injected into http/https based urls
+	u, err := url.Parse(sourceRepository)
+	if err != nil {
+		return "", fmt.Errorf("unable to parse provided gitUrl")
+	}
+	cred := url.URL{}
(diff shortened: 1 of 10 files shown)


cmd/template_gitcredential.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/uselagoon/build-deploy-tool/internal/helpers"
+	"github.com/uselagoon/build-deploy-tool/internal/lagoon"
+)
+
+var templateGitCredential = &cobra.Command{
+	Use:     "git-credential",
+	Aliases: []string{"gitcred", "gc"},
+	Short:   "Create a git credential file",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		gitURL, err := TemplateGitCredential("/home/.git-credentials")
```
bomoko, line 19: So just two things here
1 - I'd store the `/home/.git-credentials` in a const
2 - I don't know if it makes sense (I think it does) but it might make sense to have this as an argument that could be passed in - that way we can explicitly say in `build-deploy.sh` where this store file should end up, and it's not just a magic string stored between the calling code and the golang

bomoko: for (1) I'm thinking that's the default, stored in a const.

bomoko (changes requested):

cmd/template_gitcredential.go
shreddedbacon: I'll change it to a flag then, pretty simple addition.

cmd/template_gitcredential.go
```diff
@@ -0,0 +1,119 @@
...
+		return "", fmt.Errorf("unable to parse provided gitUrl")
+	}
+	// if this is a http or https based url, it may need a username and password
+	if helpers.Contains([]string{"http", "https"}, u.Scheme) {
+		// only create the git credential if the source repo is a http/https repostiory
+		urls, err := configureGitCredentials(lagoonYAML, variables)
+		if err != nil {
+			return "", err
+		}
+		var s string
```
bomoko, line 57: This could/should be replaced with a stringbuilder usually.
I think you could also stream directly to the file as well - but a stringbuilder would be more efficient in terms of memory allocation.

This isn't a trainsmash, given the amount of data/mem we're speaking about here, just a comment and doesn't necessarily rise to a blocking issue.

internal/generator/helpers_generator.go
```diff
@@ -186,3 +187,17 @@ func determineRefreshImage(serviceName, imageName string, envVars []lagoon.Envir
...
+
+// GetLagoonEnvVars will get the project and environment variables from the build
+// unmarshal them and then merge them down, where environment will override project
+func GetLagoonEnvVars() []lagoon.EnvironmentVariable {
+	projectVariables := helpers.GetEnv("LAGOON_PROJECT_VARIABLES", "", false)
+	environmentVariables := helpers.GetEnv("LAGOON_ENVIRONMENT_VARIABLES", "", false)
+	// unmarshal and then merge the two so there is only 1 set of variables to iterate over
+	projectVars := []lagoon.EnvironmentVariable{}
+	envVars := []lagoon.EnvironmentVariable{}
+	_ = json.Unmarshal([]byte(projectVariables), &projectVars)
```
bomoko, line 199: any reason not to catch/check these errs?

bomoko (approved):

PR #427, opened 2025-03-25
Feature: Adds environment type to insights config maps

bomoko:
Extends insights labels to include the environment type (development/production) in the insights config maps.


shreddedbacon (approved):
Yep makes sense. You'll need to update the branch before merge though.

PR #428, opened 2025-03-27
refactor: buildkit failed to get layer

shreddedbacon:
Another case for #413


tobybellwood (approved):
yeah, ugh, mhhh - another one to add to the list, looks good!

PR #429, opened 2025-04-03
refactor: add apex domain for wildcard ingress

shreddedbacon:
This adds support for adding the apex domain of a wildcard route to the ingress. This will be enabled by default, but can be disabled if necessary by adding `wildcardApex: false` to the route in `.lagoon.yml`
```
environments:
  main:
    routes:
      - node:
          - example.com:
              tls-acme: false
              wildcard: true
              wildcardApex: false
```


rocketeerbkw (approved):
Tested with `LAGOON_ROUTES_JSON` locally and it's working.

PR #432, opened 2025-04-24
fix: missing support for `mongo` service types

rocketeerbkw:
The [Lagoon docs](https://docs.lagoon.sh/concepts-advanced/service-types/#mongo) list the mongo service types as: `mongo`, `mongo-single`, and `mongo-dbaas` but the build-deploy-tool only has services types using `mongodb*`.

Attempting to deploy a `lagoon.type: mongo-single` would fail because the deployment would never get templated, and the build would fail because the deployment is missing.

There is a service map but it was missing most of the types. This PR adds all possible mongo types and adds tests for all the mapped types.


shreddedbacon (approved):
Nice, thanks!

PR #434, opened 2025-05-19
refactor: disable automount with admin flag override

shreddedbacon:
Refactor the automount service account token to allow for administratively enabling it via lagoon-remote only.

It is still disabled by default, and only the `ADMIN_LAGOON_FEATURE_FLAG_AUTOMOUNT_SERVICE_ACCOUNT_TOKEN=true` variable being added to the `lagoon-remote` or remote-controller will be able to enable this functionality.


tobybellwood (approved):
Quick and simple - local test OK

PR #437, opened 2025-05-22
feat: enable persistent storage in redis 8+ Lagoon images

rocketeerbkw:
Starting in uselagoon/redis-8, there is no longer a separate persistent image, and an env var needs to be added to enable peristent storage.

https://github.com/uselagoon/lagoon-images/pull/1296


tobybellwood:
Nice, thanks! Eventually, we might even be able to deprecate the entirely separate images that have only a variable different :facepalm:

shreddedbacon (approved):

shreddedbacon:
This is fairly harmless, going to merge.

PR #438, opened 2025-05-28
Configure AppSec in CI

smlx:
Partially addresses #333


smlx:
I think this is good to go. There is some complexity in the four build workflows but hopefully the change is correct. The PR docker image build worked, so that is a good sign :D

shreddedbacon:
@tobybellwood looping you in for some of the changes make to the build workflows as I'm not fully across these

smlx:
Disabled the dockerhub attestation and rebased.

PR #439, opened 2025-06-10
feat: add support for external service type

shreddedbacon:
# Description

This introduces a new service type called `external`. When defined, this type won't provision a deployment with a running container, instead it will create a service with an `ExternalName` record that resolves to the provided environment, or specific domain record.

Additionally, network policy support has been added to `.lagoon.yml` to allow for defining which organization(s), project(s), or environment(s) are able to access a service in the defined environment. By default all Lagoon environments are locked down so that no other environment can access them. These policies are used to grant access.

By combining these two features, users can deploy a service in one environment and apply a network policy that permits access from other environments. In another environment, they can then configure an external service type to point to the shared service.

One caveat is that this communication is done internally within a cluster, so this can only work if all workloads are on the same cluster. Using the `domain` field in an external service could potentially be used for cross cluster, but this domain would need to be able to be resolved, and reachable in some way. It may not cover all use cases, and may not work in all scenarios.

# Usage
## External Service Type
Defining an external service type is done in the docker compose file. An example of this is as follows
```yaml
services:
  myservice:
    image: ...
    labels:
      lagoon.type: external
      lagoon.external.service: '{"name":"basic","project":"other-project","environment":"main"}'
```
> [!IMPORTANT]
> Other options like image, or build context will be ignored by Lagoon. They may need to be defined for local development to work, but if you don't need the service locally, you could look at using [profiles](https://docs.docker.com/compose/how-tos/profiles/) to prevent them from starting locally.

Breaking this down, we have defined a service called `myservice`, and the `lagoon.type` label is set to `external`. This tells Lagoon to check the `lagoon.external.service` label for where to configure this service to point to. This data is a JSON representation of the following:

* `name` (required) - The name of the service in the other environment you want to communicate with
* `project` (required) - The name of the project that the service exists in
* `environment` (required) - The name of the environment as it is in Lagoon (eg, main, feature/branch, pr-123), not the namespace or truncated safe name.
* `domain` (optional) - The domain record to use. If this is defined, the other 3 fields are ignored.

> [!IMPORTANT]
> When referencing an external service from another project or environment, you need to ensure you use the correct port, which will be the port that is exposed on the service. This is generally the port that has been defined in the docker-compose file for the service you're wanting to reference. Eg, if an `nginx` service exposes port `8080`, this is the port you'll need to use.

## Network Policies
Using network policies allows a user to define which `organizations` and/or `projects` can access a service in the environment.

Using the following options, a user can grant all projects from an organization, or using other fields reduce the scope. For example, limiting only to production environments, or excluding certain projects.
```yaml
network-policies:
- service: myservice
  organizations:
  # this allows any environment from any project in example-org1 to access myservice
  - name: example-org1
  # this allows only production environments from any project in example-org1 to access myservice
  - name: example-org2
    environment-type: production
  # this allows any environment from any project except project-a in example-org1 to access myservice
  - name: example-org3
    exclude-projects:
      - name: project-a
```

If allowing an organization is too much, a user can also allow specific projects, environments, environment types, or exclude environments, or exclude pullrequest environments.
```yaml
network-policies:
- service: myservice
  projects: 
  # this allows any environment from example-project1 to access myservice
  - name: example-project1
  # this allows only the main environment from example-project2 to access myservice
  - name: example-project2
    environment: main
  # this allows only the production environment from example-project3 to access myservice
  - name: example-project3
    environment-type: production
  # this allows any environment except feature/develop from example-project4 to access myservice
  - name: example-project4
    exclude-environments:
    - name: feature/develop
  # this allows any environment except pullrequest environments from example-project5 to access myservice
  - name: example-project5
    exclude-pullrequests: true
```

> [!IMPORTANT]  
> It is possible to define a policy that only allows certain environments in the organizations section (for example only production environments). Then by defining a project policy in the projects section, you can selectively allow other environments to access it from more specific project override, for example allowing more than production from one or more projects that you want to be able to also access that service.
>
> This also means that some policies may not be possible to, so a user may have to define everything under projects to be more specific, rather than using the organization section.

### Defining policies
Network policies can be defined in three ways in a `.lagoon.yml` file. 

* At the root of the `.lagoon.yml` file under `network-policies`
* Under an environment under `network-policies`, similar to how routes would be defined. If defined here, the root is ignored
* Under an environment, but within a polysite setup. If defined here, others are ignored.

#### At the root
```yaml
network-policies:
- service: myservice
  organizations: 
  - name: example-org
```

#### Under a environment
```yaml
environments:
  main:
    network-policies:
    - service: myservice
      organizations: 
      - name: example-org
```

#### Under a polysite environment
```yaml
example-project:
  environments:
    main:
      network-policies:
      - service: myservice
        organizations: 
        - name: example-org
```

### DIFF
--- a/cmd/template_lagoonservices.go
+++ b/cmd/template_lagoonservices.go
@@ -145,6 +145,20 @@ func LagoonServiceTemplateGeneration(g generator.GeneratorInput) error {
 		}
 		helpers.WriteTemplateFile(fmt.Sprintf("%s/isolation-network-policy.yaml", savedTemplates), templateBytes)
 	}
+	serviceNetPols, err := servicestemplates.GenerateServiceNetworkPolicies(*lagoonBuild.BuildValues)
+	if err != nil {
+		return fmt.Errorf("couldn't generate template: %v", err)
+	}
+	for _, serviceNetPol := range serviceNetPols {
+		templateBytes, err := servicestemplates.TemplateNetworkPolicy(&serviceNetPol)
+		if err != nil {
+			return fmt.Errorf("couldn't generate template: %v", err)
+		}
+		if g.Debug {
+			fmt.Printf("Templating networkpolicy manifests %s\n", fmt.Sprintf("%s/networkpolicy-%s.yaml", savedTemplates, serviceNetPol.Name))
+		}
+		helpers.WriteTemplateFile(fmt.Sprintf("%s/networkpolicy-%s.yaml", savedTemplates, serviceNetPol.Name), templateBytes)
+	}
 	return nil
 }
 
--- a/cmd/template_lagoonservices_test.go
+++ b/cmd/template_lagoonservices_test.go
@@ -627,6 +627,38 @@ func TestTemplateLagoonServices(t *testing.T) {
 			templatePath: "testoutput",
 			want:         "internal/testdata/basic/service-templates/test-basic-single-k8upv2",
 		},
+		{
+			name:        "test-basic-external-service",
+			description: "tests a basic deployment with a basic and an external service and network policy to share a service",
+			args: testdata.GetSeedData(
+				testdata.TestData{
+					ProjectName:     "example-project",
+					EnvironmentName: "main",
+					Branch:          "main",
+					LagoonYAML:      "internal/testdata/basic/lagoon.externalservice.yml",
+					ImageReferences: map[string]string{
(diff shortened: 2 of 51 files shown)


bomoko (changes requested):
Is there a way we can flag this as the user using an experimental feature if they use this? Just as a build warning?

bomoko (approved):
This is really great. LGTM

PR #441, opened 2025-07-07
fix: multiline cronjob validation

jnsalsa:
# Problem

The current `ValidateCronjob` func checks for `\n` characters in the command, but does so after trimming the leading and trailing whitespace characters.

https://github.com/uselagoon/build-deploy-tool/blob/3d00d84607cda413c304cbf2183312c472ae7927/cmd/validate_lagoonyml.go#L87-L96

This does not catch commands that end in a `\n` and will fail the cronjob with the error:  `syntax error: unterminated quoted string`

Checking crontab shows the issue:

```
> cat /lagoon/crontabs/crontab 
* * * * * flock -n /tmp/cron.lock/xxxxxx 'command
'
```

This isn't caught because the following test cases are currently allowed as apparent single lines, but they are invalid due to the trailing \n at the end inserted by the clip:

https://github.com/uselagoon/build-deploy-tool/blob/3d00d84607cda413c304cbf2183312c472ae7927/internal/testdata/validate-lagoon-yml/cronjobs/singleline-cronjobs.lagoon.yml#L31-L37

### DIFF
--- a/cmd/validate_lagoonyml.go
+++ b/cmd/validate_lagoonyml.go
@@ -85,11 +85,9 @@ func init() {
 // ValidateCronjob returns an error if the command for the cronjob has any
 // newlines, and nil otherwise.
 func ValidateCronjob(c *lagoon.Cronjob) error {
-	command := strings.TrimSpace(c.Command)
-
-	if strings.Contains(command, "\n") {
+	if strings.Contains(strings.TrimSpace(c.Command), "\n") || strings.HasSuffix(c.Command, "\n") {
 		return fmt.Errorf("invalid cronjob, multiline commands are not supported: %q",
-			command)
+			c.Command)
 	}
 
 	return nil
--- a/internal/testdata/validate-lagoon-yml/cronjobs/multiline-cronjobs.lagoon.yml
+++ b/internal/testdata/validate-lagoon-yml/cronjobs/multiline-cronjobs.lagoon.yml
@@ -32,17 +32,26 @@ environments:
           multiline
           command
 
-      - name: block scalar folded clipped
+      - name: block scalar folded clipped blank newline
         command: >
           multiline
 
           command
 
-      - name: block scalar folded stripped
+      - name: block scalar folded stripped blank newline
         command: >-
           multiline
 
           command
+
+      - name: block scalar folded clipped double
+        command: >
+          singleline
+          command
+
+      - name: block scalar folded clipped single
+        command: >
+          singleline command
 
       - name: block scalar folded keep
         command: >+
--- a/internal/testdata/validate-lagoon-yml/cronjobs/singleline-cronjobs.lagoon.yml
+++ b/internal/testdata/validate-lagoon-yml/cronjobs/singleline-cronjobs.lagoon.yml
@@ -28,13 +28,6 @@ environments:
         command: |-
           singleline command
 
-      - name: block scalar folded clipped 1
-        command: >
-          singleline
-          command
-      - name: block scalar folded clipped 2
-        command: >
-          singleline command
       - name: block scalar folded stripped 1
         command: >-
           singleline


cmd/validate_lagoonyml.go
```diff
@@ -85,11 +85,9 @@ func init() {
 // ValidateCronjob returns an error if the command for the cronjob has any
 // newlines, and nil otherwise.
 func ValidateCronjob(c *lagoon.Cronjob) error {
-	command := strings.TrimSpace(c.Command)
-
-	if strings.Contains(command, "\n") {
+	if strings.Contains(strings.TrimSpace(c.Command), "\n") || strings.HasSuffix(c.Command, "\n") {
```
rocketeerbkw, line 88: The root cause of the error is that the validator trimmed the command string, but the templating does not. I think a more accurate test case would be:

```suggestion
	if strings.Contains(c.Command, "\n") {
```

The alternative option would be to not change the validator, and instead trim the command when templating inpod and native cronjobs. This would allow users to use the `>` yaml syntax to have a more readable cronjob definition and prevent newlines from being used.

rocketeerbkw (changes requested):

cmd/validate_lagoonyml.go
shreddedbacon: Are you going to also update the template component?

jnsalsa: @shreddedbacon

I've had a quick look.  It might be best to update the generator instead of the templates.  

https://github.com/jnsalsa/build-deploy-tool/commit/857e696d53dbf4e35cb726943459488550c95f7a#diff-65faa4119ec92040bd5c7fd2340103c542310e8e6c846c2ff676333917a2fa51L528-R531

jnsalsa: Thanks, just updated again.

rocketeerbkw (commented):
> I believe your suggested change to the validator, and an update the templater that trims the trailing whitespace would ensure best possible outcome here.

My suggestion was to change the validator OR update the template logic, not do both. The important part I wanted to point out in my review was that it's a problem if the validation step and template step treat the command string differently. Now the template trims whitespace, but the validator doesn't.

We have a release very soon and are aware that this is impacting production sites, so we will discuss internally if this will get merged and do a follow up for the above, or if it needs to be addressed before merging. You can leave it, and we'll get back to you.

jnsalsa:
At this point, from a usability perspective, the validator shouldn't change - otherwise there would be an influx of suddenly "invalid" build configurations.  What we want instead is to accomodate the "invalid" builds.

I've reverted the change to the validator to the original logic with one minor addition.  The error log now shows the raw command instead of the trimmed command which will be useful for debugging.

Thanks for helping get this organised.

shreddedbacon:
You should run `make test` to test locally. You've got failures now. This probably won't make this current release at the moment.

internal/testdata/validate-lagoon-yml/cronjobs/singleline-cronjobs.lagoon.yml
```diff
@@ -28,13 +28,6 @@ environments:
         command: |-
           singleline command
```
jackwrfuller, line 30: I think theses lines should be added back in.
```suggestion
      - name: block scalar folded clipped 1
        command: >
          singleline
          command
      - name: block scalar folded clipped 2
        command: >
          singleline command
```

internal/testdata/validate-lagoon-yml/cronjobs/multiline-cronjobs.lagoon.yml
```diff
@@ -32,18 +32,27 @@ environments:
...
           command
 
-      - name: block scalar folded stripped
+      - name: block scalar folded stripped blank newline
         command: >-
           multiline
 
           command
 
+      - name: block scalar folded clipped double
```
jackwrfuller, line 47: I think these lines should be removed.

jackwrfuller (changes requested):
I believe the issue here is that 
```yaml
      - name: block scalar folded clipped 1
        command: >
          singleline
          command
      - name: block scalar folded clipped 2
        command: >
          singleline command
```
are both valid single line commands since they are parsed as "singeline command\n", so should not be included in `multiline-cronjobs.lagoon.yml` since this file is a list of _invalid_ cron commands.

internal/testdata/validate-lagoon-yml/cronjobs/multiline-cronjobs.lagoon.yml
jnsalsa: The original plan was to make them invalid as they are _technically_ multiline commands.  However, given the latest comments I believe the approach is now to allow the commands to be valid by trimming the whitespace.

So, it makes sense to move them back to the other file.

PR #444, opened 2025-07-11
fix: cronjob refactoring and improvements

### LINKED ISSUE #443: Cronjob in pod vs native k8s check improvements
The current [check](https://github.com/uselagoon/build-deploy-tool/blob/main/internal/helpers/helpers_cron.go#L156) for if a cronjob runs in a pod or as a native k8s job doesn't work properly for all cases.

Need to refactor to properly check if the cronjob is less than 30 minutes that it goes into the pod instead of native k8s

jackwrfuller:
Resolves #443 

Passes all tests and introduces new ones to cover new functionality. Has about 73% coverage, and I manually inspected the missing coverage - mostly error pathways not explicitly checked for in tests, which I judge to not be a huge issue.

Also refactors the existing code around cronjobs to better suit the new code.

## Background

Lagoon supports a superset of the cron specification by allowing developers to include cron definitions in the `.lagoon.yml` file.
A cronjob can be implemented as either a job run in a pod, or as a kubernetes-native job.
The scheduling of kubernetes-native job is resource intensive, so for cronjobs that occur 'frequently' it is preferable to have them run in a pod.
Conversely, it is wasteful to build, deploy, and keep alive a whole pod for a job that occurs 'infrequently'.
Lagoon requires a way to decide how to implement a cronjob, and therefore must make a normative judgement on what constitutes 'frequently'.
This pull request defines a frequency metric for cronjob schedules and uses it in comparison with a 'frequency ceiling' to make a more precise decision on cronjob implementation.

## Current Implementation

The current decision is made by the following function:

```go
func IsInPodCronjob(schedule string) (bool, error) {
	splitSchedule := strings.Split(strings.Trim(schedule, " "), " ")
	// check the provided cron splits into 5
	if len(splitSchedule) == 5 {
		for idx, val := range splitSchedule {
			if idx == 0 {
				match1, _ := regexp.MatchString("^(M|H|\\*)/([0-5]?[0-9])$", val)
				if match1 {
					// A Minute like M/15 (or H/15 or */15 for backwards compatibility) is defined, create a list of minutes with a random start
					// like 4,19,34,49 or 6,21,36,51
					params := getCaptureBlocks("^(?P<P1>M|H|\\*)/(?P<P2>[0-5]?[0-9])$", val)
					step, err := strconv.Atoi(params["P2"])
					if err != nil {
						return false, fmt.Errorf("cron definition '%s' is invalid, unable to determine minutes value", schedule)
					}
					if step < 30 {
						return true, nil
					}
				}
				match2, _ := regexp.MatchString("^\\*$", val)
				if match2 {
					// this runs every minute
					return true, nil
				}
			}
		}
	}
	return false, nil
}
```

This function attempts to capture the frequency, albeit in a far from complete manner, and compares it 30 - i.e it is trying to pose the question: does this cronjob run approximately every 30 minutes or less? 
The key issue is that this function does not consider many of the possible schedules allowed by the Cron specification.
For example:

```
"2-59/10 * * * *"
```

This job runs approximately every 10 minutes, so should definitely be run in a pod as staked by the "30 minute" rule.

The specification also allows for the following:

```
"0/30,15/30 * * * *"
```

Despite the two `x/30`, this job actually run every 15 minutes and should therefore by run in a pod.
While the above two examples are 'gotchas' in the sense of `IsInPodCronjob`, they are obvious as to what 'frequency' means in this context.
However, consider the following schedule:

```
"1,59 * * * *"
```

What is the frequency of this job? 
Well, a job runs at the 1st and 59th minute every hour.
Therefore, there is a gap in time of 58 minutes between them, right?
Wrong, because we are actually doing modular arithmetic here, and there is also the 2 minute gap between minute 58 and 1.
Again though, it is fairly obvious what to do: take the average, giving an answer of 30 minutes.
Obviousness breaks down however in the face of the general specification, and so we must have a general methodology for deciding how to calculate 'frequency' for a given cron schedule.

## New Approach

My approach relies on the following insight: a cron schedule does not constrain periods of time greater than months, and so repeats yearly.
Every cron schedule defines the minutes of the year for which the command should be run, and therefore we can represent a cronjob as a subset of the integers modulo 525,960 (the number of minutes in a year).
We can therefore also define the 'frequency' of a cronjob as some metric over the set of distances between the elements of this set.
I further propose that for this metric we choose the _median_, as a desirous property of this metric for us is robustness (that is, insensivity to outliers).

This pull request introduces the following general algorithm for calculating cron frequency:

1. First, it 'standardises' a cron schedule, which I define to mean _derandomisation_: schedules with "M" or "H" are replaced with actual numbers.

2. Second, each part of the schedule is 'normalised', which I define to mean replacing the various syntaxic options for specifying times (e.g `10-15/5` or `*/15` or `5,12-40/3`) with a comma separated list of times. For example, `"0-45/15"` becomes `"0,15,30,45"`. 

3. Next, these normalised fields are 'flattened' into one big list representing all the minutes in a year. For example, `"0,30 0,1,2 * * *"` becomes `[0, 30, 60, 90, 120, 150]`

4. Then distances are calculated from the flattened list. In the previous example, the distances would be `[30, 30, 30, 30, 30, 525810]` with the final distance being that between 150 and 525960. 

5. Finally, the median is taken of these distances. Unsurprisingly, it is 30 minutes. If we had taken the average instead, it would be much greater than 30.

Except that this is a lie. For optimisation purposes, if you only care about the "30 minutes or less", then the vast majority of cases are answered by only considering minutes and hours.
So my implementation ignore the other 3 fields.
Now while there would be some edge cases that this would introduce, I judge that they would be negligible for our purposes.
If we feel that we want to cover _all_ possibilities, then we could extend the pull request to fully implement the above algorithm.


shreddedbacon:
You might need to update this to check if the other three fields are not `*`.

I added a test for the following, which should return `false`, but it returns true. While the minutes are within the required metric to put it into a pod, the day of the month being defined immediately nukes that.
```
0,15,30,45 * 4 * *
```

Basically, if any of the last 3 fields of the cronjob are not `*`, then it won't run in a pod.

internal/cron/cron.go
```diff
@@ -0,0 +1,500 @@
...
+	if cjTimeout > time.Duration(24*time.Hour) {
+		return fmt.Errorf("timeout for cronjob %s cannot be longer than 24 hours", cj.Name)
+	}
+
+	return nil
+}
+
+func calculateScheduleMetric(schedule string) (int, error) {
+	if schedule == "" {
+		return 0, fmt.Errorf("Schedule cant be empty")
```
shreddedbacon, line 267: Minor thing. Once we turn on linting in CI, errors that start with uppercase characters will report as an issue. If you can fix this up to be lowercase. There is another one just below this one too, double check for others and fix them up if you see them.

internal/generator/backups.go
```diff
@@ -92,7 +95,7 @@ func generateBackupValues(
 		}
 	}
 
-	buildValues.Backup.BackupSchedule, err = helpers.ConvertCrontab(buildValues.Namespace, newBackupSchedule)
+	buildValues.Backup.BackupSchedule, err = cron.StandardizeSchedule(newBackupSchedule, cksum.Cksum([]byte(fmt.Sprintf("%s\n", buildValues.Namespace))))
```
shreddedbacon, line 98: Rather than have the cksum called here, could you handle the cksum change in the `StandardizeSchedule` function so the argument passed in is just the namespace name?

shreddedbacon (changes requested):

internal/cron/cron.go
jackwrfuller: Have fixed for all my code, but I note that there are some outstanding cases of this preexisting:
Screenshot 2025-07-21 at 11 47 29 am

Would you like me to create a 'chore' issue to address this?

internal/generator/backups.go
jackwrfuller: Completed.

internal/cron/cron_test.go
```diff
@@ -0,0 +1,457 @@
...
+import (
+	"reflect"
+	"sort"
+	"strings"
+	"testing"
+
+	mapset "github.com/deckarep/golang-set/v2"
+)
+
+func ptr(b bool) *bool {
```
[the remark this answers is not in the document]
jackwrfuller: Modified to use `BoolPtr()`. Also refactored `BoolPtr()` to be simpler.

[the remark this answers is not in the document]
shreddedbacon: Missed this one https://github.com/jackwrfuller/build-deploy-tool/blob/cronjob-refactor/internal/generator/services_test.go#L16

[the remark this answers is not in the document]
jackwrfuller: oopsies, mb. Fixed.

shreddedbacon:
@jackwrfuller can you please update this branch with the latest changes from `main` please

shreddedbacon (approved):
Since none of the previous test results needed to be modified, I'm happy for this to merge. While the change in the logic to determine the interval is more complex, it does capture more jobs that would benefit from being run inside a long running pod. And its still basically the same 30 minutes logic.

I ran this through a few local environments with cronjobs to ensure everything still deployed correctly too, just for good measure.

Will may need to come up with a short ELI5 blurb/info block that we could add to the [documentation](https://docs.lagoon.sh/concepts-basics/lagoon-yml/#environment-cronjobs) to describe the behaviour.

PR #446, opened 2025-07-28
Fixes deprecated call, adds slightly more debugging info

bomoko:
We've experienced a few instances where a post-rollout task seems to hang. This is _very_ infrequent, and difficult to replicate. However, the function we're using to exec in the pods has been deprecated and it would be useful to have _slightly_ more info about the running of the post-rollout task itself.
This PR addresses both of these.


shreddedbacon (approved):
This is a good starting point. Hopefully changing out the deprecated call is the only thing that needed to be done. Since the reported issue isn't currently reproducible though, we'll have to wait and see.

PR #447, opened 2025-08-11
Minimal error handing for GenerateDeploymentTemplate

bomoko:
A failure in `generatePodTemplateSpec` in the wild was causing a nil reference and the BDT was crashing mid deploy because of unhandled errors from `generatePodTemplateSpec`.
This simply deals with the error.

### DIFF
--- a/internal/templating/templates_cronjob.go
+++ b/internal/templating/templates_cronjob.go
@@ -119,7 +119,10 @@ func GenerateCronjobTemplate(
 				if err != nil {
 					return nil, err
 				}
-				podTemplateSpec, _ := generatePodTemplateSpec(buildValues, serviceValues, serviceTypeValues, cronjob.ObjectMeta, templateAnnotations, nCronjob.Name, nCronjob.Command)
+				podTemplateSpec, err := generatePodTemplateSpec(buildValues, serviceValues, serviceTypeValues, cronjob.ObjectMeta, templateAnnotations, nCronjob.Name, nCronjob.Command)
+				if err != nil {
+					return nil, err
+				}
 				// end cronjob template
 				cronjob.Spec.JobTemplate.Spec.Template = *podTemplateSpec
 				result = append(result, *cronjob)


shreddedbacon (changes requested):
You'll need to also update this to return the error https://github.com/uselagoon/build-deploy-tool/blob/chore-add-error-handling-for-generateDeploymentTemplate/internal/templating/templates_cronjob.go#L122

bomoko:
> You'll need to also update this to return the error https://github.com/uselagoon/build-deploy-tool/blob/chore-add-error-handling-for-generateDeploymentTemplate/internal/templating/templates_cronjob.go#L122

Sweet, @shreddedbacon, your wish, my command (you know the drill).
I know you mentioned something about wanting to shift around the place where we add logging info into the returned errors - so we can do this if you'd like too (I just couldn't recall which you wanted where).
Although I think this works as is as well.

shreddedbacon:
> > You'll need to also update this to return the error [`chore-add-error-handling-for-generateDeploymentTemplate`/internal/templating/templates_cronjob.go#L122](https://github.com/uselagoon/build-deploy-tool/blob/chore-add-error-handling-for-generateDeploymentTemplate/internal/templating/templates_cronjob.go#L122)
> 
> Sweet, @shreddedbacon, your wish, my command (you know the drill). I know you mentioned something about wanting to shift around the place where we add logging info into the returned errors - so we can do this if you'd like too (I just couldn't recall which you wanted where). Although I think this works as is as well.

Yeah, I was going to suggest moving the error wrapping out of the `generatePodTemplateSpec` so that it returned just errors, but after looking through it, I think it is fine to leave the error wrapping where it is. So the changes here are fine now :rocket:

shreddedbacon (approved):

ocean:
Hi there Lagoonists, I got this error (in the GovCMS Lagoon instance) about a week ago, dug around in the BDT code, and then came here to send in an MR for it, and now I see you've already fixed it, woo!

My change added an error log message (and tests) though, here it is if you like 😁 

```go
podTemplateSpec, err := generatePodTemplateSpec(buildValues, serviceValues, serviceTypeValues, deployment.ObjectMeta, templateAnnotations, serviceTypeValues.PrimaryContainer.Name, "")
if err != nil {
  return nil, fmt.Errorf("couldn't generate deployment template for service %s: %v", serviceValues.OverrideName, err)
}
```

shreddedbacon:
> My change added an error log message (and tests) though, here it is if you like 😁

Have you got a branch with the tests coverage? That might be useful to add to this.

PR #449, opened 2025-08-15
feat: align default values for autogen/custom routes insecure behavior

rocketeerbkw:
The default value for `insecure` is currently different for autogenerates routes and custom routes. This PR changes the default value of autogenerated routes to be the same as custom routes, `Redirect`.


rocketeerbkw:
This might be a BC break we don't want to make? I discovered the discrepancy while updating the docs in https://github.com/uselagoon/lagoon/pull/3969.

shreddedbacon:
> This might be a BC break we don't want to make? I discovered the discrepancy while updating the docs in [uselagoon/lagoon#3969](https://github.com/uselagoon/lagoon/pull/3969).

Yeah, not sure. I guess maybe need to do some checks to see how many of our users are explicitly setting it to redirect because of this.

If the majority are doing it, then maybe it isn't that big of a deal. Most users could be handling the redirect elsewhere too (CDN layer eg..)?

PR #451, opened 2025-08-25
chore: refactor backup templ gen

### LINKED ISSUE #425: refactor prebackpod template generation
the prebackupod template generator is using an old method of seeding the template.

this should be updated to be similar to the newer service templates

jackwrfuller:
# Description

Closes #425 

This PR refactors the generation and serialisation of the prebackup pod templates.

Due to the lack of control over how the k8s-sig yaml package chooses to parse strings, I had to modify the tests to accommodate how it wanted to represent the backup command string. I dont believe the command itself has changed, but best to double check that.

A major change has been that I did not use k8upv1alpha1 pod specs. I had intended to have separate generation functions for the two versions, but as it turned out the only difference that I could find was in the Type meta, and all the tests seem to pass no problems only referencing k8upv1 pod specs.

I delved into the two version's struct definitions, and I dont see any differences for this particular case. There may be something I am missing, but it seems fine?


internal/templating/template_prebackuppod_new_test.go
shreddedbacon, line 1: Why the rename?

jackwrfuller: oops, forgot to rename

shreddedbacon (changes requested):
Only one thing picked up.

Luckily the schemas for both types are currently the same, but if they do end up changing in the future, [consolidating](https://github.com/uselagoon/build-deploy-tool/pull/451/files#diff-062683c30e40ef3b68289b4c6fd39397d836d3a5f2d24e68fe43923f03372083R93-R105) them down so that only the v1 CR is used may be problematic.

jackwrfuller:
> Only one thing picked up.
> 
> Luckily the schemas for both types are currently the same, but if they do end up changing in the future, [consolidating](https://github.com/uselagoon/build-deploy-tool/pull/451/files#diff-062683c30e40ef3b68289b4c6fd39397d836d3a5f2d24e68fe43923f03372083R93-R105) them down so that only the v1 CR is used may be problematic.

How would you like me to proceed now then? Do you want me to keep as is, or separate the function out into a v1 and v2?

shreddedbacon:
> > Only one thing picked up.
> > Luckily the schemas for both types are currently the same, but if they do end up changing in the future, [consolidating](https://github.com/uselagoon/build-deploy-tool/pull/451/files#diff-062683c30e40ef3b68289b4c6fd39397d836d3a5f2d24e68fe43923f03372083R93-R105) them down so that only the v1 CR is used may be problematic.
> 
> How would you like me to proceed now then? Do you want me to keep as is, or separate the function out into a v1 and v2?

Fine for now, ideally we will phase v1 out entirely at some point.

PR #457, opened 2025-09-15
refactor: docker-compose validation to support alternative label style

shreddedbacon:
Refactors the docker compose validation commands to read the docker compose file from the .lagoon.yml file, and check it exists. This surfaces if someone hasn't added their docker-compose file to the repository, as previously it would fail with a obscure error.

Additional refactoring on the way data from the docker compose file is used in the build scripts to not need the use of `yq` to check labels. This fixes an issue where if labels on docker-compose services are defined in an [alternative](https://github.com/uselagoon/build-deploy-tool/blob/core-v2.27.0/internal/testdata/basic/docker-compose.yml#L10-L12) way no longer fail.


tobybellwood (approved):
Tested working, nice refactor

PR #461, opened 2025-09-23
fix: typos

jackwrfuller:
# Description

Fixes typos :)


shreddedbacon (approved):
Most of the reference changes in the `json` tags are not exposed externally, so this is fine.

PR #465, opened 2025-10-07
feat: support for build hook entrypoints

shreddedbacon:
This adds support for build hooks. This is where an operator of a Lagoon can customise their build image to add hooks that can perform additional functionality during a build.

This is designed mainly as a way to augment a build, rather than perform modifications of resources provisioned during a build. Modifying resources provisioned by the build tool outside of the build tool has the potential to cause unexpected or unknown issues.

Hooks are executed in the directory in order, it is recommended to prefix your hooks with incrementing numbers like `01-hook1`, `02-hook2`, etc...

The hooks expect one of the following return codes from any hook scripts used.
0: :white_check_mark: all good.
1: :x: failure will fail the build
2+: :warning: will flag the step with a warning, but the build will continue

Adding scripts needs to be done when building the dockerfile, for example, creating the directory for the hook, and then adding the executable into that directory.
```
RUN mkdir -p /kubectl-build-deploy/hooks/pre-lagoon-yaml-validation
COPY test.sh /kubectl-build-deploy/hooks/pre-lagoon-yaml-validation/01-test.sh
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -94,6 +94,7 @@ RUN architecture=$(case $(uname -m) in x86_64 | amd64) echo "amd64" ;; aarch64 |
 
 RUN mkdir -p /kubectl-build-deploy/git
 RUN mkdir -p /kubectl-build-deploy/lagoon
+RUN mkdir -p /kubectl-build-deploy/hooks
 
 WORKDIR /kubectl-build-deploy/git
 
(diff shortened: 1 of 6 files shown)


rocketeerbkw (approved):
Tested a simple `pre-lagoon-yaml-validation` hook. Was able to trigger failure, warning, and success. Logs were output correctly.

PR #468, opened 2025-10-12
refactor: refresh the token before patching step

shreddedbacon:
When a build runs for a long time, the projected service account token consumed at the start of the build may expire, causing issues with steps that happen after it has expired.

This small change is to refresh the token before it is used in each step.


bomoko (approved):
Makes sense.

PR #471, opened 2025-11-17
feat: support for optional revision history setting

### LINKED ISSUE #470: Allow customizing of revision history limit
The revision history limit for Kubernetes deployments is by default 10 [1]. While this allows going back to review old states of a deployment, it is not always required and can consume a large amount of resources over time. Introducing a customizable default value for`.spec.revisionHistoryLimit` at the cluster level should be good enough.

[1] https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/deployment-v1/#DeploymentSpec

shreddedbacon:
Adds support for an admin settable environment variable on a remote-controller, this allows for configuring a lower than the default value of 10 for the deployments replicaset history

```
ADMIN_LAGOON_FEATURE_FLAG_DEPLOYMENT_REVISION_HISTORY=5
```

closes #470

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -395,19 +395,19 @@ func NewGenerator(
 	if buildValues.Resources.Limits.Memory != "" {
 		err := ValidateResourceQuantity(buildValues.Resources.Limits.Memory)
 		if err != nil {
-			return nil, fmt.Errorf("provided memory limit %s is not a valid resource quantity", buildValues.Resources.Limits.Memory)
+			return nil, fmt.Errorf("provided memory limit %s is not a valid resource quantity, contact your Lagoon administrator", buildValues.Resources.Limits.Memory)
 		}
 	}
 	if buildValues.Resources.Limits.EphemeralStorage != "" {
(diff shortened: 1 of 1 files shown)


internal/generator/generator.go
```diff
@@ -377,6 +377,15 @@ func NewGenerator(
 	}
 
 	// check admin features for resources
+	revisionHistory := CheckAdminFeatureFlag("DEPLOYMENT_REVISION_HISTORY", false)
+	if revisionHistory != "" {
+		rhInt, err := strconv.Atoi(revisionHistory)
+		if err != nil {
+			return nil, fmt.Errorf("revision history does not convert to integer, contact your Lagoon administrator")
```
rocketeerbkw: Aren't lagoon administrators the only ones that can set this value?

rocketeerbkw:
```suggestion
			return nil, fmt.Errorf("provided revision history %s is not a valid integer", revisionHistory)
```

Aren't lagoon administrators the only ones that can set this value?

rocketeerbkw (approved):

PR #472, opened 2025-12-09
feat: traefik support

shreddedbacon:
This adds support for traefik middleware as a replacement for ingress-nginx.

It supports reading of specific ingress-nginx annotations that may be provided by the `.lagoon.yml` and converts them to traefik middleware as part of the deployment process.

# Supported annotation migrations

We're supporting a limited set of `ingress-nginx` annotations and handling converting them to compatible traefik middleware. A list of those annotations is provided:

* `nginx.ingress.kubernetes.io/whitelist-source-range` and `nginx.ingress.kubernetes.io/denylist-source-range` - will convert into traefik middleware to accomplish the same task
* `nginx.ingress.kubernetes.io/permanent-redirect` and `nginx.ingress.kubernetes.io/temporal-redirect` will convert into traefik middleware that does a regex replace on the nginx value into one that should work with traefik
* `nginx.ingress.kubernetes.io/server-snippet` will handle converting `set_real_ip_from`, `add_header`, and `more_clear_headers` into the traefik counterparts. anything else will be ignored.
* `nginx.ingress.kubernetes.io/enable-cors`, `nginx.ingress.kubernetes.io/cors-allow-origin`, `nginx.ingress.kubernetes.io/cors-allow-methods`, `nginx.ingress.kubernetes.io/cors-allow-headers`, `nginx.ingress.kubernetes.io/cors-allow-credentials`, `nginx.ingress.kubernetes.io/cors-expose-headers`, `nginx.ingress.kubernetes.io/cors-max-age` - if `enable-cors` is provided, other annotations values will be converted into the traefik counterpart

# Restricted annotations

All traefik prefixed annotations that might be provided by a user in the `.lagoon.yml` are rejected as invalid and will fail a build. This is a protection mechanism. As annotations are evaluated, we may expose their configuration options directly via `.lagoon.yml` options, rather than via direct annotations.

# Idling support
Support for environment idling is also built into `aergia-controller` and will be released prior to changes. 
> now released in v0.6.0 of aergia-controller

# Migrations
As we progress closer to migrations, more information will be provided about how Lagoon operators can handle migrations. A separate migration tool will also be made available that will perform the above migrations to traefik on environments without requiring a build to take place. This way Lagoon operators will be able to handle preparations for migrating easier.

The build-deploy-image will also offer a mode that will allow builds to pre-create traefik middlewares and add annotations to environments, prior to changing an ingress class. This also offers a better pre-migration experience.

Setting `LAGOON_FEATURE_FLAG_DEFAULT_TRAEFIK_MIDDLEWARE=enabled` in the `remote-controller` can be used to enable Traefik support in the build. Traefik CRDs must be installed in clusters for this mode to be used, otherwise build failures will be encountered.


internal/generator/middlewares.go
```diff
@@ -0,0 +1,281 @@
...
+
+			headers := serverSnippetAddHeader(value)
+			if len(headers) > 0 {
+				// add any headers from the server-snippet to the custom response headers
+				routeHeaders.CustomResponseHeaders = headers
+				containsHeaders = true
+			}
+		}
+		// handle cors settings
+		if value, ok := route.Annotations["nginx.ingress.kubernetes.io/enable-cors"]; ok && value == "true" {
```
bomoko, line 193: I'm assuming the answer is "no" here, but is there any issue with enabling cors without any of the other cors annotations? Since this can fall through without raising a warning or setting anything at all. 
This works in the other direction as well, but obviously if you don't actually enable cors, whatever annotations might be set won't do anything.

I don't think this is an actual real issue - just wondering (since there's no warning or anything)

bomoko: leaving this here as proof I read this. Other issues discussed outside of thread.

bomoko (approved):
This PR has many moving parts, although it has been tested a number of times in situ, and has been run through a number of times by hand as well as with with LLM assistance.

Happy to approve with the caveat that there are some changes (for instance, the x-lagoon header) that clearly cannot be replicated one to one between ingress controller types.

bomoko (approved):
New changes make sense.

PR #477, opened 2026-02-03
fix: always check `global` env vars combined with `build` or `runtime`

rocketeerbkw:
This fixes two annoyances around the use of Lagoon env var feature flags.

The first issues is a few places in the code that looks up environment variables treats the `global` scope like a third type. There is never a case where the `global` scope should be checked alone, it should always be combined with `build` or `runtime`. And vice-versa, checking for `runtime` or `build` scoped env vars should also always check in `global`.

The second is some feature flag value comparisons happen in bash which was done in a case sensitive way.


shreddedbacon (approved):
Makes sense

PR #478, opened 2026-02-18
Removes Insights from BDT

bomoko:
This PR deprecates the insights step on the BDT - it is replaced by a post build insights step that is controlled exclusively by the insights-remote controller (https://github.com/uselagoon/insights-remote/pull/103)


legacy/build-deploy-docker-compose.sh
```diff
@@ -1895,39 +1895,40 @@ if [ "${LAGOON_VARIABLES_ONLY}" != "true" ]; then
...
-    set +e # Ensure failures in exec-generate-insights-configmap.sh don't halt the entire build
-    INSIGHTS_WARNING_COUNT=0
-    for IMAGE_NAME in "${!IMAGES_BUILD[@]}"
-    do
-      IMAGE_TAG="${IMAGE_TAG:-latest}"
-      IMAGE_FULL="${REGISTRY}/${PROJECT}/${ENVIRONMENT}/${IMAGE_NAME}:${IMAGE_TAG}"
-      insightsOutput=$(. /kubectl-build-deploy/scripts/exec-generate-insights-configmap.sh 2>&1)
-      if (exit $?); then
-        echo "${insightsOutput}"
+    if [ "$(featureFlag INSIGHTS_EXTERNAL | tr '[:upper:]' '[:lower:]')" = disabled ]; then
```
shreddedbacon: I think change to `!=` so that if the flag is not set (like it currently isn't anywhere) then the in-pod insights runs
```suggestion
    if [ "$(featureFlag INSIGHTS_EXTERNAL | tr '[:upper:]' '[:lower:]')" != enabled ]; then
```

Then to enable the external, you have to set the flag.

shreddedbacon (changes requested):

shreddedbacon:
Closing, as this isn't required after re-evaluating.

After https://github.com/uselagoon/lagoon-charts/pull/677 was merged, the flag to turn external pod scanning with `enableBuildScanning` was set to true. This controls the external build pod scanning

The `lagoonFeatureFlagDefaultInsights` value controls the injection of the variable `LAGOON_FEATURE_FLAG_DEFAULT_INSIGHTS` which runs the post deploy insights in the build (https://github.com/search?q=repo%3Auselagoon%2Flagoon-charts%20lagoonFeatureFlagDefaultInsights&type=code). Since this is a feature flag, and is disabled by default, it can simply be removed if users are using insights. We should update the charts to remove the `lagoonFeatureFlagDefaultInsights` value from injection for newer versions. But leave the support for the `LAGOON_FEATURE_FLAG_DEFAULT_INSIGHTS` in the build-deploy-tool for at least the next release for backwards compatibility reasons.

PR #483, opened 2026-03-05
fix: correctly set and consume dbaas for variables only deployments

### LINKED ISSUE #482: Variable only deployments do not contain all databases from dbaas
Spotted today on a site that has 2 databases defined in `docker-compose.yml`.

After a variables only deployment, it seems that only the second database credentials are in the running pods

```
[XXX]production@cli-drupal:/app$ env | sort | grep -i database
SIMPLESAML_DATABASE=XXX_iPhlS
```

After a full deploy, this restored it

```
[XXX]production@cli-drupal:/app$ env | sort | grep -i database
MARIADB_DATABASE=XXX_WtjSJ
SIMPLESAML_DATABASE=XXX_iPhlS
```

I would consider this a critical issue.

shreddedbacon:
closes #482

### DIFF
--- a/legacy/build-deploy-docker-compose.sh
+++ b/legacy/build-deploy-docker-compose.sh
@@ -1379,37 +1379,43 @@ if [ "${LAGOON_VARIABLES_ONLY}" != "true" ]; then
   # standard deployment
 else
   # variable only deployment
+  MARIADB_CONSUMER_COUNT=0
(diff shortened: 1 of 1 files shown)


bomoko (approved):
Logic is solid.

bomoko (approved):

PR #484, opened 2026-03-19
Update tools and images to latest compatible versions

tobybellwood:
This pull request updates the `Dockerfile` to use newer base images and tool versions, improving security and compatibility. The main changes are version bumps for Alpine, Docker, Go, kubectl, Helm, and yq.

The most contentious change here is the move to docker 29.3.0 - it will need a more thorough test (and may warrant upgrading the docker golang package to match)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,12 @@ ARG UPSTREAM_TAG
 ARG GO_VER
 FROM ${UPSTREAM_REPO:-uselagoon}/commons:${UPSTREAM_TAG:-latest} AS commons
 
-FROM golang:${GO_VER:-1.25}-alpine3.22 AS golang
+FROM golang:${GO_VER:-1.25}-alpine3.23 AS golang
 
 RUN apk add --no-cache git
-RUN go install github.com/a8m/envsubst/cmd/envsubst@v1.4.2
+# renovate: datasource=github-releases depName=a8m/envsubst
+ENV ENVSUBST_VERSION=v1.4.3
+RUN go install github.com/a8m/envsubst/cmd/envsubst@${ENVSUBST_VERSION}
 
 WORKDIR /app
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -71,19 +71,19 @@ ENV TMPDIR=/tmp \
...
 
 # Defining Versions
-ENV KUBECTL_VERSION=v1.30.3 \
-    HELM_VERSION=v3.15.3
+ENV KUBECTL_VERSION=v1.35.2 \
+    HELM_VERSION=v3.20.1
 
 RUN apk add -U --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing aufs-util \
-    && apk upgrade --no-cache openssh openssh-keygen openssh-client-common openssh-client-default \
+    && apk upgrade --no-cache \
```
shreddedbacon: Were these packages hardcoded for a reason? I remember in #181 about the reversion, but was there a reason to keep them at all?

tobybellwood: Removing the specified packages from this line doesn't affect their being upgraded (or not), so they're safe to remove from explicit upgrade requests. There was a time when the upstream Docker mismatched Alpine (I think).

Dockerfile
```diff
@@ -71,19 +73,25 @@ ENV TMPDIR=/tmp \
...
+# renovate: datasource=github-tags depName=kubernetes/kubernetes
+ENV KUBECTL_VERSION=v1.35.3
+# renovate: datasource=github-releases depName=helm/helm
+ENV HELM_VERSION=v3.20.2
+# renovate: datasource=github-releases depName=mikefarah/yq
+ENV YQ_VERSION=v4.52.5
+# renovate: datasource=github-releases depName=uselagoon/lagoon-linter
+ENV LAGOON_LINTER_VERSION=v0.8.0
 
 RUN apk add -U --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing aufs-util \
```
shreddedbacon: i don't think we need to install `aufs-util` anymore either. i don't see anything that requries it. docker might have needed it in the past, but it no longer uses it

shreddedbacon: https://gist.github.com/shreddedbacon/8fb4c1c55e2d2a358f11081dc509e7be

Two builds from this branch. One as it currently is, one with `aufs-util` not being installed.

Both worked without issues.

shreddedbacon:
> The most contentious change here is the move to docker 29.3.0 - it will need a more thorough test (and may warrant upgrading the docker golang package to match)

I did some deploys in a local-stack and didn't hit any issues with this upgrade, so seems like it is probably fine?

shreddedbacon (approved):
Everything seems to work as expected

PR #485, opened 2026-03-23
refactor: resource removal wait and additional info changes

shreddedbacon:
Just a small refactor to wait for resources to be removed before continuing. This allows for the resources to be removed before the final state collection runs to ensure that the latest state is saved into the services config map

Additional changes to volume information too to reflect changes coming in the API to support volumes on environments.


CGoodwin90 (approved):
Tested locally alongside https://github.com/uselagoon/lagoon/pull/4065 and works as expected - didn't encounter any issues.

PR #498, opened 2026-04-24
Updates redis backup to only copy non-temp root rdb files

bomoko:
This PR updates the backup command only backup root rdp files, rather than the entire directory - this should prevent backup failures when the AOF files are being written to.


shreddedbacon (approved):
Does what it says on the tin

PR #511, opened 2026-07-15
fix: error check for buildkit layer failures

shreddedbacon:
Updates to docker appear to have changed the error presented when a buildkit layer cache is no longer present or encounters an issue.

Previously it was
```
ERROR: failed to solve: failed to prepare
```
Now it appears to be
```
ERROR: failed to build: failed to solve: failed to prepare
```


bomoko (approved):
makes sense and preserves past behavior

PR #515, opened 2026-07-23
fix: just check if url is http/https for git credentials

shreddedbacon:
Just a small fix to git url credential check that the url only has prefix for http/https. 

If a git url is an SCP style git url, eg `git@example.com:lagoon-demo.git` vs `ssh://git@example.com/lagoon-demo.git`, it will currently produce an error in the build log `Error: unable to parse provided gitUrl` which can mislead users that there is a problem


bomoko (approved):
makes sense.

I don't suppose it makes sense to no have this as an "error" in the logs.

